PR #25, opened 2018-02-19
Move full nvticache to Redis 

kroosec:


cfi-gb:
Cross-ref to the scanner changes: https://github.com/greenbone/gvm-libs/pull/25

PR #35, opened 2018-03-06
Use "nvticache10" value to identify cache KB.

kroosec:
So that an nvtcache from openvas-9 isn't wrongly used. Also store the
feed version in that key.

### DIFF
--- a/util/nvticache.c
+++ b/util/nvticache.c
@@ -41,11 +41,14 @@
 #include <sys/stat.h> /* for stat, st_mtime */
 #include <time.h>     /* for time, time_t */
 #include <stdlib.h>     /* for atoi */
+#include <stdio.h>     /* for fopen */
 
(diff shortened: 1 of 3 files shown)


util/nvticache.c
```diff
@@ -41,12 +41,15 @@
...
 #include <time.h>     /* for time, time_t */
 #include <stdlib.h>     /* for atoi */
+#include <stdio.h>     /* for fopen */
 
 #include "kb.h" /* for kb_del_items, kb_item_get_str, kb_item_add_int */
 
 #undef  G_LOG_DOMAIN
 #define G_LOG_DOMAIN "lib  nvticache"
 
+#define NVTICACHE_STR "nvticache10"
```
bjoernricks: NVTICACHE_STR is defined twice. In the c and the header file.

PR #45, opened 2018-03-21
Fix gpgme/gcrypt include path detection

tuxmaster:
Use the include directory from gpgme-config/libgcrypt-config.


bjoernricks:
Thanks

PR #82, opened 2018-05-04
2.23.0 of uuidlib is ok to compile it

tuxmaster:
Tested on CentOS7

### DIFF
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -46,7 +46,7 @@ pkg_check_modules (GIO REQUIRED gio-2.0>=2.42)
 pkg_check_modules (GNUTLS REQUIRED gnutls>=3.2.15)
 
 # for uuidutils we need uuidlib
-pkg_check_modules (UUID REQUIRED uuid>=2.25.0)
+pkg_check_modules (UUID REQUIRED uuid>=2.23.0)
 
 # for sshutils we need libssh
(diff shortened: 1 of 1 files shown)


wiegandm:
As the version number of `libuuid` actually represents the version of the `util-linux` release containing it looks like the API is comparatively stable.

At a first glance, the changes between 2.23 and 2.25 look to be related only to documentation:
  https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/Documentation/releases/v2.25-ReleaseNotes#n815

Changes between 2.25 and 2.29 on the other hand seem to mostly bugfixes, with no direct API changes as far as I can tell:
  https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/Documentation/releases/v2.26-ReleaseNotes#n662
  https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/Documentation/releases/v2.27-ReleaseNotes#n404
  https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/Documentation/releases/v2.28-ReleaseNotes#n417

wiegandm:
So I think it should be safe for downstream packagers to apply patches downgrading the dependencies in this case.

But especially for the `master` branch using more recent library versions is a sensible approach IMHO.

kroosec:
I agree with @wiegandm on having such specific requirements changed downstream. Master branch is still months away from hitting a stable release anyway.

PR #105, opened 2018-07-18
Add host string matching functions

timopollmeier:
The new functions gvm_host_in_hosts_str and gvm_hosts_str_contains have
been added to check if a single host or host string matches a string
defining multiple hosts without having to convert the latter to a
gvm_hosts_t.

### DIFF
--- a/base/hosts.c
+++ b/base/hosts.c
@@ -827,6 +827,42 @@ gvm_get_host_type (const gchar *str_stripped)
 }
 
 /**
+ * @brief Creates a new gvm_vhost_t object.
+ *
+ * @param[in] value     Vhost value.
+ * @param[in] source    Source of hostname.
+ *
+ * @return Pointer to new vhost object.
+ */
+gvm_vhost_t *
+gvm_vhost_new (char *value, char *source)
+{
+  gvm_vhost_t *vhost;
+
+  vhost = g_malloc0 (sizeof (gvm_vhost_t));
+  vhost->value = value;
+  vhost->source = source;
+
+  return vhost;
+}
+
+/**
+ * @brief Frees the memory occupied by an gvm_vhost_t object.
+ *
+ * @param[in] vhost Vhost to free.
+ */
+static void
+gvm_vhost_free (gpointer vhost)
+{
+  if (vhost)
+    {
+      g_free (((gvm_vhost_t *) vhost)->value);
+      g_free (((gvm_vhost_t *) vhost)->source);
+    }
+  g_free (vhost);
+}
+
+/**
  * @brief Creates a new gvm_host_t object.
  *
  * @return Pointer to new host object, NULL if creation fails.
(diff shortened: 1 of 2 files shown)


base/hosts.c
```diff
@@ -1749,6 +1807,193 @@ gvm_host_in_hosts (const gvm_host_t *host, const struct in6_addr *addr,
...
+
+  if (host == NULL)
+    return 0;
+
+  ret = gvm_host_in_hosts_str (host, hosts_str);
+  gvm_host_free (host);
+
+  return ret;
+}
+
```
kroosec: Actually, the whole patch could be discarded by using something like this in GVM:

{
gvm_hosts_t *hosts = gvm_hosts_new (hosts_str);
gvm_hosts_t *hosts2 = gvm_hosts_new (find_host_str);
if (!hosts || !hosts2 || hosts2->count != 0)
{
gvm_hosts_free (hosts);
gvm_hosts_free (hosts2);
return 0
}

int ret = gvm_host_in_hosts (hosts2->hosts->data, NULL, hosts);
gvm_hosts_free (hosts);
gvm_hosts_free (hosts2);
return ret;
}

base/hosts.c
```diff
@@ -897,6 +897,64 @@ gvm_host_free (gpointer host)
...
+          g_free (stripped_host_str);
+          return NULL;
+        }
+    }
+
+  g_free (stripped_host_str);
+
+  return host;
+}
+
```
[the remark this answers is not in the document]
timopollmeier: I wanted to avoid creating a gvm_hosts_t but in this case it shouldn't be too much overhead if all that's passed in should be single IP adresses.

base/hosts.h
```diff
@@ -152,6 +152,12 @@ int
 gvm_host_in_hosts (const gvm_host_t *, const struct in6_addr *,
                    const gvm_hosts_t *);
 
+int
+gvm_host_in_hosts_str (const gvm_host_t *host, const char *hosts_str);
```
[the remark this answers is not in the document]
timopollmeier: You're right. It's also not really useful without a way to generate a single gvm_host_t.

base/hosts.c
timopollmeier: I wanted to avoid larger allocations that would occur for ranges and CIDR blocks since the function could be called for a large number of results.

kroosec: Hi @timopollmeier I did some more tests (with tests/test-hosts) to see where the hosts module is i.r.w memory allocations and speed. In order to get a really big memory footprint / speed impact, one has to use large host numbers (ie. millions of hosts) which would have the same impact everywhere else (scans etc,.)

I believe that adding 200LOC that pretty much duplicates existing functionality instead of using a couple lines shouldn't be done, unless there is a really good reason for it.

timopollmeier: @kroosec, I've updated the manager PR to use the existing libraries function like you suggested, with one change: I've used gvm_hosts_new_with_max(...) to avoid issues with large allocation in case a large hosts range is passed to the function.

timopollmeier: I'll be closing this PR now since it's no longer necessary with the new solution for GVM.

PR #121, opened 2018-09-18
Backport scanner PR #171.

kroosec:
Fix parameter pollution in certain nasl functions.

Some NASL functions use get_var_by_name() instead of
get_local_var_by_name(), which "climbs" in the context when searching
for a variable ie. if a parameter 'foo' is not provided, the function
would use the variable 'foo' if found.

Unify the behaviour by replacing all get_local_var_by_name()
with get_var_by_name() calls, and change the later to not climb
in the nasl context when searching for the variable.

### DIFF
--- a/nasl/nasl_cert.c
+++ b/nasl/nasl_cert.c
@@ -798,7 +798,7 @@ nasl_cert_query (lex_ctxt *lexic)
     }
 
   /* Get the index which defaults to 0.  */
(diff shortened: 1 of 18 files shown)


kroosec:
To reproduce:

hostname = 'foo';
source = 'bar';
add_host_name();
display(get_host_name() + '|' + get_host_name_source());

cfi-gb:
Direct ref for the scanner PR: https://github.com/greenbone/openvas-scanner/pull/171

PR #152, opened 2018-11-22
Switch KB from using unsorted sets to lists.

kroosec:
This ensures ordering of a key's inserted values, and removes unicity of
values for a key.

PR can't be merged at the moment, as many VTs out there still rely on the unicity "feature/bug"

Related post from 2014, on usage of unordered sets instead of lists: https://marc.info/?l=openvas-development&m=139326215027633&w=2


kroosec:
The 2nd commit of the PR handles the values unicity case, and adds a warning, which would make finding VTs that rely on this behaviour easier.

PR #167, opened 2019-01-16
Fix lost of path to config file

jjnicola:
After sending -SIGHUP to the scanner to force a reload,
the path to the config file was overwritten. Now the configurations
in the config file are preserved.
Thanks  @kroosec for the patch improvement.


jjnicola:
Fixes greenbone/openvas-scanner#213

PR #171, opened 2019-01-18
Add the test to CMakeList to be executed by ctest.

jjnicola:
Now make tests finds the test.


bjoernricks:
Maybe https://libcheck.github.io/check/ can be used for test assertions.

PR #176, opened 2019-01-22
Solve TODO. Remove gvm_auth_tear_down().

jjnicola:
It was an empty function with a reference to a list to be freed, which was removed
in 2015. So, nothing to be freed. There is no plan to extend it. Therefore remove it.


jjnicola:
This function was used by gvmd. To avoid conflicts, the PR greenbone/gvmd#339 should be merged.

PR #181, opened 2019-01-30
Contributing notes.

jjnicola:
Move section related to development from INSTALL.md to the contributing section in README.md.
Add subsection "Code style and formatting".


bjoernricks:
> Looks good!
> 
> I think the `README.md` file is a good place for this information right now. If the 'contributing' section becomes larger, please consider moving it to a separate file like `CONTRIBUTING.md`.

I've advised using the README file. IMHO it's better to put all stuff in a bigger README then using different files. Maybe we should discuss this topic someday in more detail.

PR #184, opened 2019-02-03
Memleak

kroosec:
Fix miscellaneous memory leaks.


kroosec:
> I guess this should also be merged into the new release branch right?

@bjoernricks Yes, we also backport https://github.com/greenbone/gvm-libs/pull/185

PR #186, opened 2019-02-04
# unsecure rpath

hsntgm:
net-analyzer/openvas: fix unsecure rpath.

Gentoo/Linux QA warns about unsecure rpaths.
It seems this patch worked for me while making package fully
compatible with Gentoo QA.

Tested by: Hasan ÇALIŞIR (hasan.calisir@psauxit.com)


hsntgm:
https://dev.gentoo.org/~zmedico/portage/doc/ch07.html#qa-scanelf-runpath

I work with v9 branch to maintain it for gentoo and your current build system stuck at this point for gentoo QA check.Maybe i am missing something.I don't say this is a issue for you just want to draw your attention.

bjoernricks:
First of all for addressing OpenVAS 9 with openvas-libraries requires a patch against the openvas-libraries-9.0 branch. Seconds the rpath is only set during build time not install time. E.g.

```sh
readelf -d /home/bricks/install/openvas-9/lib/libopenvas_* |grep -i path
```

doesn't return anything. So setting CMAKE_BUILD_WITH_INSTALL_RPATH=TRUE  would not change the rpath at install time.

hsntgm:
"First of all for addressing OpenVAS 9 with openvas-libraries requires a patch against the openvas-libraries-9.0 branch."

You mean you fixed it in 1.0 master already? Very well,  it is my bad.
I know that you develop openvas on debian and don't like package maintainers so much (e.g. arch maintainer pulled systemd services for you and you rejected him.)

I don't care at all.Just i have to open bug report and point to my ebuild about issue how i fixed it.After gentoo devops review my patches i will close pull request.

Have a nice day.

bjoernricks:
> "First of all for addressing OpenVAS 9 with openvas-libraries requires a patch against the openvas-libraries-9.0 branch."

Sorry maybe this statement was a bit rude. I was only not sure if you are aware that master branch isn't the branch you are using in gentoo for packaging OpenVAS 9.

> You mean you fixed it in 1.0 master already? Very well, it is my bad.

I am currently not sure if that is fixed already. gvm-libs is a re-write of the old openvas-libraries and have't found a line were the rpath is applied.

>I know that you develop openvas on debian

We mostly use Debian based distributions e.g. our Greenbone OS is based on Debian. But as a developer I am free of choice. Personally I am on Ubuntu. A former co-worker used Arch.

> and don't like package maintainers so much 

That's not true (at least for me). I really appreciate the hard work packagers are doing. It's always not an easy task to fix user problems and to talk to the upstream devs.

> (e.g. arch >maintainer pulled systemd services for you and you rejected him.)

I would love to integrate these patches. The rejection had political reasons. I am still in favor of merging the PR for gsad https://github.com/greenbone/gsa/pull/760 but hadn't time to discuss the issues behind the rejection again.

So back to the rpath issue. I just wanted to discuss the best solution and if the reported issue is a real issue for openvas-libraries. Currently the rpath is stripped during the installation. This is already done without your patches. I don't think setting a install rpath to the install prefix does make sense because the install prefix should be also in the linker path. So maybe it is best to remove the rpath by using [CMAKE_SKIP_RPATH](https://cmake.org/cmake/help/latest/variable/CMAKE_SKIP_RPATH.html#variable:CMAKE_SKIP_RPATH)

PR #198, opened 2019-03-17
Remove support for Redis 2.4.

kroosec:
Scanner officially supports 3.2 and higher.

See https://github.com/greenbone/openvas-scanner/pull/287


bjoernricks:
Could you please add a reason behind not supporting older versions anymore? e.g. list features not supported by these old versions or being unsupported by the devs.

janowagner:
Very welcome update. The older redis version are not tested by us anymore with old redis-versions,
so better not guess we support them and rather have 3.2 as minimum.

PR #199, opened 2019-03-20
Rework gvm_hosts_shuffle() implementation.

kroosec:
Prior implementation used an O(N^2) routine. This one has O(N) time complexity.

With a /8 network, the routine takes a couple of seconds, instead of multiple hours.


jjnicola:
It works like a charm!

PR #208, opened 2019-03-27
Backport PR #207

kroosec:
Use a dynamic array for gvm_hosts instead of doubly-linked lists.

This substantially improves handling of large sets of hosts.
ie. With a "10.0.0.0/8" hosts string, memory footprint is reduced from
1.33GB to 0.9GB. Other operations times such as creation and shuffling
are also slightly faster.

Backport of https://github.com/greenbone/gvm-libs/pull/207

### DIFF
--- a/base/hosts.c
+++ b/base/hosts.c
@@ -31,9 +31,9 @@
 
 #include "networking.h" /* for ipv4_as_ipv6, addr6_as_str, gvm_resolve */
 
-#include <arpa/inet.h>  /* for inet_pton, inet_ntop */
-#include <assert.h>     /* for assert */
-#include <ctype.h>      /* for isdigit */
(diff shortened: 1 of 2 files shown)


cfi-gb:
In https://github.com/greenbone/gvm-libs/commit/680d0c13cc85d8703bf5a5c1d262d616a799a538#commitcomment-32956396 it was noticed that the build is failing with:

```
/root/openvas-src/gvmd/gvmd/src/manage_utils.c:744:49: error: '*find_hosts->hosts' is a pointer; did you mean to use '->'?                                           
   int ret = gvm_host_in_hosts (find_hosts->hosts->data, NULL, hosts);                                                                                               
                                                 ^~                                                                                                                  
                                                 ->
```

PR #223, opened 2019-04-24
Use g_log instead of g_debug for No redis DB available message.

jjnicola:


cfi-gb:
Should also make https://github.com/greenbone/gvm-libs/issues/144 more verbose like suggested in https://github.com/greenbone/gvm-libs/issues/144#issuecomment-439536162

PR #226, opened 2019-04-30
Add kb_new_no_lock.

jjnicola:
In case there is no more available redis databases, it does not wait for available one.
Instead, it returns an error code which makes possible to identify this problem. To do this,
get_redis_ctx return an integer instead of redisContext, and receive a flag to select it
as a blocking or not blocking function.


jjnicola:
Closed this one in favor of PR #230

PR #250, opened 2019-07-21
Make QoD Type an explicit element of stuct nvti.

janowagner:
This makes it unnecessary to handle the QoD Type in the "tags" parts.

If it is added to the tags, it still will be. This is how openvas uses it.

The change is intended to support gvmd so it can use this new API element
and prevent qod_type to enter the tags element.


janowagner:
gvmd uses this API extension in https://github.com/greenbone/gvmd/pull/647

PR #251, opened 2019-07-21
Add function osp_get_vts_filtered().

janowagner:
In contrast to function get_vts(), this function allows to
call GET_VTS with the "filters" parameter.

### DIFF
--- a/osp/osp.h
+++ b/osp/osp.h
@@ -77,8 +77,12 @@ osp_get_vts_version (osp_connection_t *, char **);
 int
 osp_get_vts (osp_connection_t *, entity_t *);
 
+typedef struct {
+  char *filter; ///< the filter to apply for a vt sub-selection.
+} osp_get_vts_opts_t;
+
 int
-osp_get_vts_filtered (osp_connection_t *, const gchar *, entity_t *);
+osp_get_vts_ext (osp_connection_t *, osp_get_vts_opts_t, entity_t *);
 
 int
 osp_start_scan (osp_connection_t *, const char *, const char *, GHashTable *,
--- a/osp/osp.h
+++ b/osp/osp.h
@@ -78,6 +78,9 @@ int
 osp_get_vts (osp_connection_t *, entity_t *);
 
 int
+osp_get_vts_filtered (osp_connection_t *, const gchar *, entity_t *);
+
+int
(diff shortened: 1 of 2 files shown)


janowagner:
> I think it would be better to add an osp_get_vts_ext instead of an osp_get_vts_filtered.
> See osp_start_scan_ext for an example. This style lets us extend the api without
> changing the function definition (just the osp_get_vts_opts_t structure would need
> to be extended).

I changed it accordingly.

osp/osp.h
```diff
@@ -77,6 +77,13 @@ osp_get_vts_version (osp_connection_t *, char **);
 int
 osp_get_vts (osp_connection_t *, entity_t *);
 
+typedef struct {
+  char *filter; ///< the filter to apply for a vt sub-selection.
+} osp_get_vts_opts_t;
```
bjoernricks: Just for curiosity, why do are the typedefs put into the headers? I've seen that a lot in the past in our code base. Personally I never would make them public.

janowagner: I followed the scheme of osp_start_scan_ext().

The calling module needs to fill the "opts" properly. So it needs to know about the single
elements of it.
See here how I used it: 
https://github.com/greenbone/gvmd/pull/648/commits/691a6d2ed5705d554c11af632ff60aa191ddeea7

bjoernricks: That's exactly the issue i have with making structs public.

>get_vts_opts.filter = g_strdup_printf ("modification_time>%s", db_feed_version); 

There should be a public function like `void osp_get_vts_opts_set_filter(osp_get_vts_opts_t *opts, const char *filter)` to not let users access internals like this struct. To be clear the layout of the data in the memory should be an implementation detail and not directly accessible by users of the api.

janowagner: The scope of my patch is to aid the the patch about acceleration of the feed update.
So I applied the scheme that is done for all similar OSP and GMP API functions (about 10 of them I found) rather than introducing a second scheme for just one function.

bjoernricks: I was not asking about changing your patch. I was just curios about the reason why we make data structs public accessible. Personally I am not going to change that at all.

mattmundell: In this case we're following the scheme introduced by someone from long ago.
Can see how your suggestion would be cleaner (but more work).

PR #252, opened 2019-07-29
Use API to access nvti information.

janowagner:
The nvti struct elements should not be accessed directly.
This patch takes fixes the last occurance in gvm-libs of direct
access.

### DIFF
--- a/util/kb.c
+++ b/util/kb.c
@@ -1287,7 +1287,7 @@ redis_add_nvt (kb_t kb, const nvti_t *nvt, const char *filename)
   struct kb_redis *kbr;
   redisReply *rep = NULL;
   int rc = 0;
-  GSList *element;
+  unsigned int i;
   gchar *cves, *bids, *xrefs;
 
   if (!nvt || !filename)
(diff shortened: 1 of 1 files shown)


util/kb.c
```diff
@@ -1314,21 +1314,19 @@ redis_add_nvt (kb_t kb, const nvti_t *nvt, const char *filename)
   if (rep != NULL)
     freeReplyObject (rep);
 
-  element = nvt->prefs;
-  if (g_slist_length (element))
+  if (nvti_pref_len (nvt))
     rep = redis_cmd (kbr, "DEL oid:%s:prefs", nvti_oid (nvt));
-  while (element)
+  for (i = 0; i < nvti_pref_len (nvt); i ++)
```
mattmundell: There should not be a space before the ++.

util/kb.c
```diff
@@ -1314,21 +1314,19 @@ redis_add_nvt (kb_t kb, const nvti_t *nvt, const char *filename)
...
     {
-      nvtpref_t *pref = element->data;
+      const nvtpref_t *pref = nvti_pref (nvt, i);
 
       rep = redis_cmd (kbr, "RPUSH oid:%s:prefs %d|||%s|||%s|||%s",
-                       nvti_oid (nvt), pref->id, pref->name, pref->type,
-                       pref->dflt);
+                       nvti_oid (nvt), nvtpref_id (pref), nvtpref_name (pref),
+		       nvtpref_type (pref), nvtpref_default(pref));
```
mattmundell: There are some tabs at the beginning of the line, and there's a space missing after "nvtpref_default".

util/kb.c
janowagner: Fixed.

util/kb.c
janowagner: Fixed.

PR #253, opened 2019-07-31
Make the actual nvti struct internal.

janowagner:
This enforces to use the intended API. It
can not be bypassed anymore by accessing the elements directly.


janowagner:
https://github.com/greenbone/openvas/pull/365
needs to be merged before this one can be applied.

PR #255, opened 2019-08-07
Make solution and solution_type explicit for nvti

janowagner:
This means to add explicit methods to set and get solution and solution_type.

### DIFF
--- a/base/nvti.c
+++ b/base/nvti.c
@@ -813,6 +813,51 @@ nvti_set_solution_type (nvti_t *n, const gchar *solution_type)
 }
 
 /**
+ * @brief Add a tag to the tags of a NVT.
+ *
+ * @param n The NVT Info structure.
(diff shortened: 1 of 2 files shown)


base/nvti.c
```diff
@@ -500,9 +505,37 @@ nvti_refs (const nvti_t *n, const gchar *type, const gchar *exclude_types,
...
- * @param n The NVT Info structure of which the name should
+ * @param n The NVT Info structure of which the solution should
+ *          be returned.
+ *
+ * @return The solution string. Don't free this.
+ */
+gchar *
+nvti_solution (const nvti_t *n)
+{
+  return (n ? n->solution : NULL);
```
mattmundell: Side note: you're the only one using parens with "return".  Everywhere else uses "return 0;" style.

PR #256, opened 2019-08-14
Allow to configure the path to the redis socket

bjoernricks:
Add a new CMake directive (-DREDIS_SOCKET_PATH=/path/to/socket)to allow
to set the path to the redis unix socket at build time.

Additionally change the default path to use the default socket path of
redis on debian and ubuntu which is (/var)/run/redis/redis.sock.


bjoernricks:
> This looks OK. Not sure what you mean by "(/var)" in the description, because I don't see how /var would ever be added to the path.

Because the redis-server systemd service file uses `/var/run/...` but `/var/run` is just a symlink to `/run` nowadays :expressionless:

PR #259, opened 2019-08-22
Add function to get an osp scan status.

jjnicola:


mattmundell:
Spotted a little typo.  And the patch looks OK, but I would rather see us using the _opt_t style interface in new functions.  See osp_start_scan_ext for an example.  This gives us flexibility in the future to extend the function without changing the args.

PR #261, opened 2019-08-27
Add timestamps as explicit elements of nvti.

janowagner:


base/nvti.c
```diff
@@ -169,6 +169,9 @@ typedef struct nvti
   gchar *affected; /**< @brief Affected systems */
   gchar *impact;   /**< @brief Impact of vulnerability */
 
+  gint creation_time;     /**< @brief Time of creation, seconds since epoche */
```
mattmundell:
```suggestion
  gint creation_time;     /**< @brief Time of creation, seconds since epoch */
```

base/nvti.c
```diff
@@ -169,6 +169,9 @@ typedef struct nvti
...
+  gint modification_time; /**< @brief Time of last change, sec. since epoche */
```
mattmundell:
```suggestion
  gint modification_time; /**< @brief Time of last change, sec. since epoch */
```

base/nvti.c
```diff
@@ -464,6 +467,34 @@ nvti_impact (const nvti_t *n)
   return (n ? n->impact : NULL);
 }
 
+/**
+ * @brief Get the creation time.
+ *
+ * @param n The NVT Info structure of which the creation time should
+ *          be returned.
+ *
+ * @return The creation time in seconds since epoche.
```
mattmundell:
```suggestion
 * @return The creation time in seconds since epoch.
```

base/nvti.c
```diff
@@ -464,6 +467,34 @@ nvti_impact (const nvti_t *n)
...
+  return (n ? n->creation_time : 0);
+}
+
+/**
+ * @brief Get the modification time.
+ *
+ * @param n The NVT Info structure of which the modification time should
+ *          be returned.
+ *
+ * @return The modification time in seconds since epoche.
```
mattmundell:
```suggestion
 * @return The modification time in seconds since epoch.
```

base/nvti.c
mattmundell: I think these should be time_t instead of gint.  That's what we use in gvmd, and it matches the return from C's time().

janowagner: resolved.

base/nvti.c
janowagner: Done.

base/nvti.c
janowagner: Done.

base/nvti.c
janowagner: Done.

base/nvti.c
janowagner: Done.

PR #262, opened 2019-09-02
Add osp_get_performance()

jjnicola:
Related to PR greenbone/ospd#131

### DIFF
--- a/osp/osp.c
+++ b/osp/osp.c
@@ -444,6 +444,53 @@ osp_delete_scan (osp_connection_t *connection, const char *scan_id)
   return ret;
 }
 
+/**
+ * @brief Get performance graphics from an OSP server.
+ *
+ * @param[in]   connection  Connection to an OSP server.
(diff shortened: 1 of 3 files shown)


osp/osp.c
```diff
@@ -444,6 +444,53 @@ osp_delete_scan (osp_connection_t *connection, const char *scan_id)
   return ret;
 }
 
+/**
+ * @brief Get performance graphics from an OSP server.
+ *
+ * @param[in]   connection  Connection to an OSP server.
+ * @param[in]   opts        Struct containing the options to apply.
+ * @param[out]  graph       Graphic base64 encoded.
+ * @param[out]  error       Pointer to error, if any.
```
mattmundell:
```suggestion
 * @param[out]  error       Either NULL or a pointer for an error string.  Only set if return is 1.
 *                          Caller must free.
```

mattmundell: Actually this is consistent with the rest of the file.  I will update them all at once later.

PR #263, opened 2019-09-06
Add nvti_add_tag() for adding a single tag to tags.

janowagner:
So far it was only possible to set the entire tags
string. Now it is possible to append a single tag.

### DIFF
--- a/base/nvti.c
+++ b/base/nvti.c
@@ -1070,7 +1070,7 @@ nvti_set_solution_type (nvti_t *n, const gchar *solution_type)
 int
 nvti_add_tag (nvti_t *n, const gchar *name, const gchar *value)
 {
-  char *new;
+  char *newtag;
 
   if (!n)
     return (-1);
@@ -1083,14 +1083,14 @@ nvti_add_tag (nvti_t *n, const gchar *name, const gchar *value)
 
   if (n->tag)
     {
-      new = g_strconcat (n->tag, "|", name, "=", value, NULL);
+      newtag = g_strconcat (n->tag, "|", name, "=", value, NULL);
       g_free (n->tag);
     }
   else
-    new = g_strconcat (name, "=", value, NULL);
+    newtag = g_strconcat (name, "=", value, NULL);
 
-  if (new && new[0])
-    n->tag = new;
+  if (newtag && newtag[0])
+    n->tag = newtag;
   else
     n->tag = NULL;
 
--- a/base/nvti.c
+++ b/base/nvti.c
@@ -1061,7 +1061,9 @@ nvti_set_solution_type (nvti_t *n, const gchar *solution_type)
  * @brief Add a tag to the NVT tags.
  *
  * @param n     The NVT Info structure.
+ *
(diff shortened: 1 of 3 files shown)


base/nvti.c
```diff
@@ -1057,6 +1057,46 @@ nvti_set_solution_type (nvti_t *n, const gchar *solution_type)
   return (0);
 }
 
+/**
+ * @brief Add a tag to the NVT tags.
+ *
+ * @param n The NVT Info structure.
+ *
+ * @param name  The tag name. A copy will be created from this.
+ * @param value The tag value. A copy will be created from this.
```
mattmundell:
```suggestion

 * @param value The tag value. A copy will be created from this.
```

base/nvti.c
```diff
@@ -1057,6 +1057,46 @@ nvti_set_solution_type (nvti_t *n, const gchar *solution_type)
...
+
+  if (n->tag)
+    {
+      new = g_strconcat (n->tag, "|", name, "=", value, NULL);
+      g_free (n->tag);
+    }
+  else
+    new = g_strconcat (name, "=", value, NULL);
+
+  if (new && new[0])
```
mattmundell: This seems wrong, if you don't have other changes in mind.

new will always have an actual value due to the checks and the two g_strconcat calls above.

I think new should be declared in the if branch and the else should just assign to n->tag.

base/nvti.c
```diff
@@ -1057,6 +1057,46 @@ nvti_set_solution_type (nvti_t *n, const gchar *solution_type)
...
+ *
+ * @param name  The tag name. A copy will be created from this.
+ * @param value The tag value. A copy will be created from this.
+ *
+ * @return 0 for success. Anything else indicates an error.
+ */
+int
+nvti_add_tag (nvti_t *n, const gchar *name, const gchar *value)
+{
+  char *newtag;
```
mattmundell:
```suggestion
  gchar *newtag;
```

gchar because that's what g_strconcat returns.

janowagner: Done.

base/nvti.c
janowagner: I condensed this now.

base/nvti.c
janowagner: Hm, I rather preferred to remove the extra empty line to make it consistent.

base/nvti.c
mattmundell: Better, thanks.  newtag is only used in the 'if' branch, so you could
declare it in the 'if' branch, to make that clear.

base/nvti.c
mattmundell: Consistent with what?  :)

The rest of the file uses spacing (except for two functions right at the end).

The rest of gvm-libs does not use the spacing though.

base/nvti.c
janowagner: Done.

base/nvti.c
mattmundell: Will keep in mind for new PR.

PR #265, opened 2019-09-19
Convert incoming timestamp tags to seconds since epoch

janowagner:
OSP should only use seconds since epoch for timestamps to provide
the OSP clients with a reliable format/syntax.

This patch transforms incoming timestamps immediately.
The OpenVAS-specific parsing code for timestamp is moved here and can
later be remove from gvmd where it currently is used to parse the various
syntax's.

A positive side effect is the reduced size consumption in redis:
The string "2019-05-14 08:13:05 +0000 (Tue, 14 May 2019)"
becomes "1557821585"


jjnicola:
I can see the timestamp in the expected format in redis.
```
"last_modification=1565160945|creation_date=1131023284|cvss_base=0.0|cvss_base_vector=AV:N/AC:L/Au:N/C:N/I:N/A:N|summary=This plugin runs nmap to find open ports.|qod_type=remote_banner"```

PR #282, opened 2019-10-16
Fix sigsegv when no plugin_feed_info.inc file present.

jjnicola:
Check for null before trying to save the value in Redis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [20.4] (unreleased)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,9 +4,18 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [Unreleased]
+## [20.4] (unreleased)
 
-## [11.0.0] (unreleased)
+[11.0.1]: https://github.com/greenbone/gvm-libs/compare/gvm-libs-11.0...master
```
mattmundell: Should be something like this?

```suggestion
[20.4]: https://github.com/greenbone/gvm-libs/compare/gvm-libs-11.0...master
```

PR #285, opened 2019-10-22
Add nvti_get_tag().

janowagner:
This adds a function to get a tag value
for a certain tag name.
The main intention is to avoid routines using nvti API
needing to know about the pipe separator syntax.


base/nvti.c
```diff
@@ -763,6 +763,44 @@ nvti_tag (const nvti_t *n)
...
+ * @return The tag value string as a copy or NULL if not found.
+ *         Needs to be free'd.
+ */
+gchar *
+nvti_get_tag (const nvti_t *n, const gchar *name)
+{
+  gchar **split, **point;
+
+  if (n->tag == NULL)
+    return NULL;
```
mattmundell: There are two different styles of return in this function.  I think they should all be the same as the majority in the file.  (And a think a subsequent patch should convert the entire file to the parenthesis-less style we use everywhere else).

base/nvti.c
```diff
@@ -763,6 +763,44 @@ nvti_tag (const nvti_t *n)
...
+ *
+ * @return The tag value string as a copy or NULL if not found.
+ *         Needs to be free'd.
+ */
+gchar *
+nvti_get_tag (const nvti_t *n, const gchar *name)
+{
+  gchar **split, **point;
+
+  if (n->tag == NULL)
```
jjnicola: I would check for a null pointer here, before accessing n.
```suggestion
  if (n && n->tag == NULL)
```

PR #301, opened 2019-11-28
Add a gvm_libs_version function

bjoernricks:
Allow users of gvm-libs request the current version string.

**Checklist**:

- [x] Tests

### DIFF
--- a/base/version_tests.c
+++ b/base/version_tests.c
@@ -30,7 +30,7 @@ AfterEach (version)
 {
 }
 
-Ensure (version, compare_version)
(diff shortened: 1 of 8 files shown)


base/version_tests.c
```diff
@@ -0,0 +1,51 @@
...
+
+Describe (version);
+BeforeEach (version)
+{
+}
+AfterEach (version)
+{
+}
+
+Ensure (version, compare_version)
```
mattmundell: This should be the name of the function, followed by what it tests, so something like gvm_libs_version_compare_version or gvm_libs_versions_returns_correct_version.

PR #304, opened 2019-12-09
Fix trust and file handling for S/MIME (11.0)

timopollmeier:
The certificates used for S/MIME encryption are now added to the GPG
trust list automatically because users cannot always confirm trust.

Also, a data buffer with custom I/O functions is used to write the
encrypted data because gpgme_data_new_from_stream can cause problems
when trying to write afterwards.

**Checklist**:

- [ ] Tests


timopollmeier:
I've created a new, cleaner feature branch to be merged here: #309

PR #309, opened 2019-12-18
Fix trust and file handling for S/MIME (11.0)

timopollmeier:
The certificates used for S/MIME encryption are now added to the GPG
trust list automatically because users cannot always confirm trust.

Also, a data buffer with custom I/O functions is used to write the
encrypted data because gpgme_data_new_from_stream can cause problems
when trying to write afterwards.

**Checklist**:

- Tests N/A

### DIFF
--- a/util/gpgmeutils.c
+++ b/util/gpgmeutils.c
@@ -445,7 +445,8 @@ create_all_certificates_trustlist (gpgme_ctx_t ctx, const char *homedir)
   GString *trustlist_content;
   GError *g_err;
 
-  g_err = NULL;  gpgme_set_pinentry_mode (ctx, GPGME_PINENTRY_MODE_CANCEL);
+  g_err = NULL;
+  gpgme_set_pinentry_mode (ctx, GPGME_PINENTRY_MODE_CANCEL);
 
   trustlist_filename = g_build_filename (homedir,
                                          "trustlist.txt",
--- a/util/gpgmeutils.c
+++ b/util/gpgmeutils.c
@@ -359,6 +359,127 @@ find_email_encryption_key (gpgme_ctx_t ctx, const char *uid_email)
 }
 
 /**
+ * @brief Wrapper for fread for use as a GPGME callback.
+ *
+ * @param[in]  handle   The file handle.
+ * @param[out] buffer   The data buffer to read data into.
+ * @param[in]  size     The size of the buffer.
+ *
+ * @return The number of bytes read or -1 on error.
+ */
+static ssize_t
+gvm_gpgme_fread (void *handle, void *buffer, size_t size)
+{
+  int ret;
+  FILE *file = (FILE *)handle;
+
+  ret = fread (buffer, 1, size, file);
+  if (ferror (file))
+    return -1;
+  return ret;
+}
+
+/**
+ * @brief Wrapper for fread for use as a GPGME callback.
+ *
+ * @param[in]  handle   The file handle.
+ * @param[in]  buffer   The data buffer to read data into.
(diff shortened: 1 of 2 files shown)


util/gpgmeutils.c
```diff
@@ -358,6 +358,127 @@ find_email_encryption_key (gpgme_ctx_t ctx, const char *uid_email)
...
+ * @param[in]   file      The stream to read from and write to.
+ *
+ * @return The return value from gpgme_data_new_from_cbs.
+ */
+static gpgme_error_t
+gvm_gpgme_data_new_from_stream (gpgme_data_t *new_data, FILE *file)
+{
+  struct gpgme_data_cbs *callbacks;
+
+  callbacks = malloc (sizeof (struct gpgme_data_cbs));
```
mattmundell:
```suggestion
  callbacks = g_malloc0 (sizeof (struct gpgme_data_cbs));
```

util/gpgmeutils.c
```diff
@@ -358,6 +358,127 @@ find_email_encryption_key (gpgme_ctx_t ctx, const char *uid_email)
...
+  memset (callbacks, 0, sizeof (struct gpgme_data_cbs));
```
mattmundell:
```suggestion
```

util/gpgmeutils.c
```diff
@@ -358,6 +358,127 @@ find_email_encryption_key (gpgme_ctx_t ctx, const char *uid_email)
...
+ */
+static int
+create_all_certificates_trustlist (gpgme_ctx_t ctx, const char *homedir)
+{
+  gpgme_key_t key;
+  gchar *trustlist_filename;
+  GString *trustlist_content;
+  GError *g_err;
+
+  g_err = NULL;  gpgme_set_pinentry_mode (ctx, GPGME_PINENTRY_MODE_CANCEL);
```
mattmundell:
```suggestion
  g_err = NULL;
  gpgme_set_pinentry_mode (ctx, GPGME_PINENTRY_MODE_CANCEL);
```

util/gpgmeutils.c
```diff
@@ -358,6 +358,127 @@ find_email_encryption_key (gpgme_ctx_t ctx, const char *uid_email)
...
+gvm_gpgme_data_new_from_stream (gpgme_data_t *new_data, FILE *file)
+{
+  struct gpgme_data_cbs *callbacks;
+
+  callbacks = malloc (sizeof (struct gpgme_data_cbs));
+  memset (callbacks, 0, sizeof (struct gpgme_data_cbs));
+  callbacks->read = gvm_gpgme_fread;
+  callbacks->write = gvm_gpgme_fwrite;
+
+  return gpgme_data_new_from_cbs (new_data, callbacks, file);
```
mattmundell: Where is callbacks freed?  Should callbacks->release be set to do this?

util/gpgmeutils.c
mattmundell: We usually use the glib wrapper, which checks the malloc return.

util/gpgmeutils.c
mattmundell: Alternatively callbacks could be passed in from the caller, so it doesn't need to be freed.

util/gpgmeutils.c
mattmundell: Done in 6f4400c.

util/gpgmeutils.c
mattmundell: Done in c126b85.

util/gpgmeutils.c
mattmundell: Done in 8428f83.

PR #317, opened 2020-02-16
Add multiple severities for nvti

janowagner:
This adds a handling for severities similar to refs.

**Checklist**:

- [ ] Tests

### DIFF
--- a/base/nvti.c
+++ b/base/nvti.c
@@ -909,6 +909,31 @@ nvti_vtseverity (const nvti_t *n, guint p)
 }
 
 /**
+ * @brief Get the maximum severity score
+ *
+ * @param n The NVT Info structure.
+ *
+ * @return The severity score, -1 indicates an error.
+ */
+int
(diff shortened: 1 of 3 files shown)


timopollmeier (approved):

base/nvti.c
```diff
@@ -159,6 +159,141 @@ vtref_text (const vtref_t *r)
...
+vtseverity_new (const gchar *type, const gchar *origin, int date, int score,
+                const gchar *value)
+{
+  vtseverity_t *s = g_malloc0 (sizeof (vtseverity_t));
+
+  if (type)
+    s->type = g_strdup (type);
+  if (origin)
+    s->origin = g_strdup (origin);
+  s->date = date;
```
bjoernricks, line 202: i really think date should be optional too

bjoernricks: nevermind. it is an int and therefore can be -1 for not set or something like that.

PR #320, opened 2020-02-27
get new gvm_host_t by string

ArnoStiefvater:
Add function for getting a gvm_host_t by string. This is used in the test_alive_hosts_only feature of openvas.

Related PR:
- [openvas](https://github.com/greenbone/openvas/pull/456)
- [ospd-openvas](https://github.com/greenbone/ospd-openvas/pull/204)

### DIFF
--- a/base/hosts.c
+++ b/base/hosts.c
@@ -1537,10 +1537,10 @@ gvm_hosts_exclude (gvm_hosts_t *hosts, const char *excluded_str)
 
 /**
  * @brief Creates a new gvm_host_t from the provided host_str. host_str can
- * consist of hostnam, IPv4 address or IPv6 address.
+ * consist of hostname, IPv4 address or IPv6 address.
  *
- * @param[in] host_str The host string. A copy will be created of this within
- *                      the returned struct.
+ * @param[in] host_str The host string. A copy will be created for the returned
+ * struct.
  *
  * @return NULL if error. Otherwise, a single host structure that should be put
  * into a gvm_hosts_t structure for freeing with @ref gvm_hosts_free or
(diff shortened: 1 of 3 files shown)


base/hosts.c
```diff
@@ -1535,6 +1535,60 @@ gvm_hosts_exclude (gvm_hosts_t *hosts, const char *excluded_str)
   return gvm_hosts_exclude_with_max (hosts, excluded_str, 0);
 }
 
+/**
+ * @brief Creates a new gvm_host_t from the provided host_str. host_str can
```
mattmundell: I'd like to keep "@brief" comments to one single line.  Anything more can go in the description below.

Here I would say just "from a host string".  The description of host_str would fit better in the "@param".

base/hosts.c
```diff
@@ -1535,6 +1535,60 @@ gvm_hosts_exclude (gvm_hosts_t *hosts, const char *excluded_str)
...
+ * consist of hostname, IPv4 address or IPv6 address.
+ *
+ * @param[in] host_str The host string. A copy will be created for the returned
```
mattmundell: I don't think you need to mention that it is copied.  It is already const.

base/hosts.c
```diff
@@ -1535,6 +1535,60 @@ gvm_hosts_exclude (gvm_hosts_t *hosts, const char *excluded_str)
...
+ * freed directly via @ref gvm_host_free.
+ */
+gvm_host_t *
+gvm_host_from_str (const gchar *host_str)
+{
+  if (host_str == NULL)
+    {
+      return NULL;
+    }
+  int host_type;
```
mattmundell: I would really like to see this for variables:
 - always declared at the top of a block,
 - never initialised in the declaration, but in a new statement separated by a newline from the declarations,
 - comments about the block always come after the declarations (but before the initialisations).

Makes the code more predictable.

PR #325, opened 2020-04-20
add unit tests and small changes for port range functions

ArnoStiefvater:
Add unit tests for port range functions validate_port_range(), port_range_ranges() and port_in_port_ranges().
Allow white-space after port type specifier for consistency (it is allowed everywhere else too).
Log warning when invalid port range is used in port_range_ranges() and add assert.

**Checklist**:

- [x] Tests


base/networking_tests.c
```diff
@@ -105,6 +105,56 @@ Ensure (networking, validate_port_range)
...
+Ensure (networking, port_range_ranges)
+{
+  /* Valid port ranges. U:,T: are empty ranges which are ignored. */
+  const gchar *valid_portrange1 = "1,10-13,10-10,T:1-2,U:10-12,U:,T:";
+  assert_that (validate_port_range (valid_portrange1), is_equal_to (0));
+
+  assert_that (port_range_ranges (NULL), is_null);
+  assert_that (port_range_ranges (valid_portrange1), is_not_null);
+
+  array_t *valid_portrange1_ranges = port_range_ranges (valid_portrange1);
```
mattmundell, line 841: I'd much rather see the variables all declared at the top of the block (but keeping the definitions at the point of use).

base/networking.c
```diff
@@ -601,11 +601,19 @@ port_range_ranges (const char *port_range)
...
   array_t *ranges;
-  int tcp;
+  int tcp, err;
 
   if (!port_range)
     return NULL;
 
+  /* port_range needs to be a valid port_range string. */
+  err = validate_port_range (port_range);
+  assert (!err);
```
mattmundell, line 611: Not sure why you're asserting this.  To me the assertion says that it ret should always be 0, but because this is an exported function we can't guarantee the assertion -- the caller could provide a value that makes the assertion fail.

base/networking_tests.c
```diff
@@ -105,6 +105,56 @@ Ensure (networking, validate_port_range)
   assert_that (validate_port_range ("   T  : 1"), is_equal_to (0));
 }
 
+Ensure (networking, port_range_ranges)
+{
+  /* Valid port ranges. U:,T: are empty ranges which are ignored. */
+  const gchar *valid_portrange1 = "1,10-13,10-10,T:1-2,U:10-12,U:,T:";
```
mattmundell, line 835: Maybe add some blanks around the ':'s to test your additions?

base/networking_tests.c
```diff
@@ -103,6 +104,729 @@ Ensure (networking, validate_port_range)
...
+    ALL_TCP,
+    OPENVAS_DEFAULT,
+    ALL_IANA_ASSIGNED_TCP_AND_UDP_2020_02_12,
+    ALL_TCP_AND_NMAP_5_51_TOP_100_UDP,
+    ALL_TCP_AND_NMAP_5_51_TOP_1000_UDP,
+    NMAP_5_51_TOP_2000_TCP_AND_TOP_100_UDP,
+    WEB_SERVICES,
+  };
+
+  const char *portlists[] = {
```
mattmundell, line 123: Nice :)

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add unit tests for osp. [#315](https://github.com/greenbone/gvm-libs/pull/315)
 - Add support for test_alive_hosts_only feature of openvas. [#320](https://github.com/greenbone/gvm-libs/pull/320)
 - Add function to set and get the NVT QoD. [#321](https://github.com/greenbone/gvm-libs/pull/321)
+- Add unit tests for networking.c port list functions. [#315](https://github.com/greenbone/gvm-libs/pull/325)
```
mattmundell, line 17:
```suggestion
- Add unit tests for networking.c port list functions. [#325](https://github.com/greenbone/gvm-libs/pull/325)
```

base/networking.c
ArnoStiefvater: But if the caller provides an invalid port range then things like `"-2,10-8,90000,T:2=2,U"` are valid entries for the function. 
This means, that e.g. for the range "-2" the following unit tests pass:
```
assert_that (range->start, is_equal_to(0));
assert_that (range->end, is_equal_to(2));
```
So the user gets a `range_t` which has 3 ports (0, 1, 2) when supplying the range "-2".

Do you know an alternative for handling such cases?

mattmundell: To be clear, I'm saying that the "assert" should be removed, not that the "if (err)" should be removed.

Issue #334, opened 2020-05-13
[11.0.1] build fails, beause missing GCC option


tuxmaster5000:
The build fails with:
[ 96%] Built target gvm_base_shared
/builddir/build/BUILD/gvm-libs-11.0.1/util/xmlutils.c: In function 'find_child':
/builddir/build/BUILD/gvm-libs-11.0.1/util/xmlutils.c:1750:3: error: 'for' loop initial declarations are only allowed in C99 mode
   for (xmlNode *node = element->children; node; node = node->next)
   ^
/builddir/build/BUILD/gvm-libs-11.0.1/util/xmlutils.c:1750:3: note: use option -std=c99 or -std=gnu99 to compile your code

tuxmaster5000:
Using an newer gcc (9.1.1.) then this line will work without the patch, but later it fails with:

> cd /builddir/build/BUILD/gvm-libs-11.0.1/util && /usr/lib64/ccache/cc -DENABLE_LDAP_AUTH=1 -DENABLE_RADIUS_AUTH=1 -DGVM_LIBS_VERSION=\"11.0.1\" -DNVTICACHE_STR=\"nvticache11.0.1\" -DRADIUS_AUTH_FREERADIUS=1 -DREDIS_SOCKET_PATH=\"/run/redis/redis.sock\" -Dgvm_util_shared_EXPORTS -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/gvm-helper -I/usr/include/libxml2  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1  -m64 -mtune=generic -D_FILE_OFFSET_BITS=64 -DLARGEFILE_SOURCE=1 -g -Wall -Wextra -Werror  -fPIC   -o CMakeFiles/gvm_util_shared.dir/xmlutils.c.o   -c /builddir/build/BUILD/gvm-libs-11.0.1/util/xmlutils.c
> In file included from /usr/include/string.h:638,
>                  from /usr/include/glib-2.0/glib/gtestutils.h:30,
>                  from /usr/include/glib-2.0/glib.h:82,
>                  from /builddir/build/BUILD/gvm-libs-11.0.1/base/settings.h:30,
[...]

PR #335, opened 2020-05-13
GCC need an option for build

tuxmaster5000:
Fix
https://github.com/greenbone/gvm-libs/issues/334

### DIFF
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -134,6 +134,10 @@ if (BUILD_WITH_LDAP)
   endif (NOT LIBLDAP)
 endif (BUILD_WITH_LDAP)
 
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
+  add_compile_options(-std=c99)
+endif()
+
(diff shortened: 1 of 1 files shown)


util/CMakeLists.txt
```diff
@@ -135,7 +135,7 @@ if (BUILD_WITH_LDAP)
 endif (BUILD_WITH_LDAP)
 
 if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
```
bjoernricks, line 137:
```suggestion
if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
# set C standard for older GNU compilers (GCC >= 6.3.0)
```

tuxmaster5000: You mean < 6.3.0?

bjoernricks: > So with this patch it will work with gcc up to 6.3.0

I guess yes. I misinterpreted your comment. So for gcc >= 6.3.0 you need this patch?

tuxmaster5000: No it looks like it is only needed lower than 6.3.
I have added an version check for it,

PR #336, opened 2020-05-18
Add new scan status PENDING.

jjnicola:
**Checklist**:

- [ ] Tests

### DIFF
--- a/osp/osp.c
+++ b/osp/osp.c
@@ -572,7 +572,9 @@ osp_get_scan_status_ext (osp_connection_t *connection,
       return status;
     }
 
-  if (!strcmp (entity_attribute (child, "status"), "init"))
+  if (!strcmp (entity_attribute (child, "status"), "pending"))
+    status = OSP_SCAN_STATUS_PENDING;
+  else if (!strcmp (entity_attribute (child, "status"), "init"))
     status = OSP_SCAN_STATUS_INIT;
   else if (!strcmp (entity_attribute (child, "status"), "running"))
     status = OSP_SCAN_STATUS_RUNNING;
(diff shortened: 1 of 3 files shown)


osp/osp.c
```diff
@@ -572,6 +572,8 @@ osp_get_scan_status_ext (osp_connection_t *connection,
       return status;
     }
 
+  if (!strcmp (entity_attribute (child, "status"), "pending"))
+    status = OSP_SCAN_STATUS_PENDING;
   if (!strcmp (entity_attribute (child, "status"), "init"))
```
mattmundell, line 577:
```suggestion
  else if (!strcmp (entity_attribute (child, "status"), "init"))
```

osp/osp.h
```diff
@@ -61,6 +61,7 @@ typedef enum
 typedef enum
 {
   OSP_SCAN_STATUS_ERROR = -1, /**< Error status. */
+  OSP_SCAN_STATUS_PENDING,    /**< Init pending. */
```
mattmundell, line 64: Don't know if anything stores these, but it's safer to add to the end.

For example if gvmd was storing these states as integers in the db, then the stored states would all be wrong.

osp/osp.h
```diff
@@ -65,6 +65,8 @@ typedef enum
   OSP_SCAN_STATUS_RUNNING,    /**< Running status. */
   OSP_SCAN_STATUS_STOPPED,    /**< Stopped status. */
   OSP_SCAN_STATUS_FINISHED,   /**< Finished status. */
+  OSP_SCAN_STATUS_PENDING,    /**< Init pending. */
+
```
mattmundell, line 69:
```suggestion
```

jjnicola:
Thanks @mattmundell !

PR #339, opened 2020-05-26
Get interface used for routing to given destination

ArnoStiefvater:
Add implementation for getting the interface which is used for routing to a given destination. If provided a source address is set to the address of the interface to use.

- [X] Tests

### DIFF
--- a/base/networking.c
+++ b/base/networking.c
@@ -846,3 +846,108 @@ ip_islocalhost (struct sockaddr_storage *storage)
   return FALSE;
 }
 
+
+struct proc_entry
+{
+  gchar *interface;
+  unsigned long mask;
+  unsigned long dest;
+};
+static GSList *
+get_routes (void)
+{
+  GSList *proc_routes = NULL;
+  GError *err = NULL;
+  GIOChannel *mychannel;
+  gchar *line;
+  gchar **items_in_line;
+  int status;
+  struct proc_entry *entry;
+
+  /* Open "/proc/net/route". */
+  mychannel = g_io_channel_new_file ("/proc/net/route", "r", &err);
+  if (mychannel == NULL)
+    {
+      g_warning ("%s: %s. ", __func__,
+                 err ? err->message : "Error opening /proc/net/ipv6_route");
+      err = NULL;
+      return NULL;
+    }
+
+  /* Skip first first line of file. */
+  status = g_io_channel_read_line (mychannel, &line, NULL, NULL, &err);
+  if (status != G_IO_STATUS_NORMAL || !line || err)
+    {
+      g_warning ("%s: %s", __func__,
+                 err ? err->message
+                     : "g_io_channel_read_line() status != G_IO_STATUS_NORMAL");
+      if (line)
+        g_free (line);
+      err = NULL;
+    }
+
+  /* Until EOF or err we go through lines of file and extract Iface, Mask and
+   * Destination and but it into the to be returned list of routes.*/
+  while (1)
+    {
+      gchar *interface;
+      unsigned long mask;
+      unsigned long dest;
+      gchar *char_p;
+      int count = 0;
+
+      /* get new line */
+      status = g_io_channel_read_line (mychannel, &line, NULL, NULL, &err);
+      if ((status != G_IO_STATUS_NORMAL) || !line || err)
+        {
+          g_warning (
+            "%s: %s", __func__,
+            err ? err->message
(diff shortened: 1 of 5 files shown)


base/networking.c
```diff
@@ -846,3 +846,108 @@ ip_islocalhost (struct sockaddr_storage *storage)
...
+  while (1)
+    {
+      gchar *interface;
+      unsigned long mask;
+      unsigned long dest;
+      gchar *char_p;
+      int count = 0;
+
+      /* get new line */
+      status = g_io_channel_read_line (mychannel, &line, NULL, NULL, &err);
```
mattmundell: Looks like this is leaking the 'line' that was allocated above.

base/networking.c
```diff
@@ -762,3 +762,341 @@ ipv6_is_enabled ()
...
+ * storage_source != NULL.
+ *
+ * @return Interface name of interface used for routing to destination address.
+ * NULL if no interface found or Error.
+ */
+gchar *
+gvm_routethrough (struct sockaddr_storage *storage_dest,
+                  struct sockaddr_storage *storage_source)
+{
+  if (!storage_dest)
```
mattmundell, line 986: Putting declarations at top would be more consistent.

base/networking.c
```diff
@@ -762,3 +762,341 @@ ipv6_is_enabled ()
...
+  if (getifaddrs (&ifaddr) == -1)
+    {
+      g_debug ("%s: getifaddr failed: %s", __func__, strerror (errno));
+      return NULL;
+    }
+
+  /* IPv4. */
+  if (storage_dest->ss_family == AF_INET)
+    {
+      GSList *routes = NULL;
```
mattmundell, line 1002: This is redundant because it's defined directly below.  Recommend doing the init's after the declaration, with a line separating them, to avoid this kind of confusion.

base/networking.c
```diff
@@ -762,3 +762,341 @@ ipv6_is_enabled ()
...
+                  && (ifa->ifa_flags & (IFF_LOOPBACK)))
+                {
+                  interface_out = g_strdup (ifa->ifa_name);
+                  break;
+                }
+            }
+        }
+      else
+        {
+          struct sockaddr_in *sin_dest_p = NULL;
```
mattmundell, line 1029: Another case where it's defined, but set directly below.

base/networking.c
```diff
@@ -762,3 +762,341 @@ ipv6_is_enabled ()
...
+          for (routes_p = routes; routes_p; routes_p = routes_p->next)
+            {
+              if (((sin_dest_p->sin_addr.s_addr
+                    & ((struct route_entry *) (routes_p->data))->mask)
+                   == ((struct route_entry *) (routes_p->data))->dest)
+                  && (((struct route_entry *) (routes_p->data))->mask
+                      >= best_match))
+                {
+                  /* Interface of matching route.*/
+                  if (interface_out)
```
mattmundell, line 1050: This is not necessary.  g_free does the same check.

base/networking.c
```diff
@@ -762,3 +762,341 @@ ipv6_is_enabled ()
...
+            }
+        }
+      freeifaddrs (ifaddr);
+    }
+
+  return FALSE;
+}
+
+/** Entry of routing table /proc/net/route */
+struct route_entry
```
mattmundell, line 857: Just an idea, would be convenient to add rout_entry_t, because you're casting often to this type below.  We usually do it that way, at least on gvmd side.

base/networking.h
```diff
@@ -113,4 +113,8 @@ port_in_port_ranges (int, port_protocol_t, array_t *);
 int
 ipv6_is_enabled ();
 
+gchar *
+gvm_routethrough (struct sockaddr_storage *storage_dest,
```
mattmundell, line 117: Should not have the arg names.

base/networking.c
```diff
@@ -762,3 +762,340 @@ ipv6_is_enabled ()
...
+  /* Until EOF or err we go through lines of file and extract Iface, Mask and
+   * Destination and put it into the to be returned list of routes.*/
+  while (1)
+    {
+      gchar *interface, *char_p;
+      unsigned long mask, dest;
+      int count = 0;
+
+      /* Get new line. */
+      status = g_io_channel_read_line (file_channel, &line, NULL, NULL, &err);
```
mattmundell, line 915: This still looks like a leak to me.  A line is allocated by the g_io_channel_read_line called before the while, but that line is not freed before the call to g_io_channel_read_line here.

PR #342, opened 2020-06-04
Move alive detection module Boreas into gvm-libs

ArnoStiefvater:
- [x] Tests


ArnoStiefvater:
Does no preserver git history of. New PR will be made.

PR #349, opened 2020-06-12
Simplify scan restrictions

ArnoStiefvater:
Remove max_alive_hosts scan restriction as it is currently not planned to be used and makes the code more complex.

**Checklist**:

- [ ] Tests
- [ ] [CHANGELOG](https://github.com/greenbone/gvm-libs/blob/master/CHANGELOG.md)

### DIFF
--- a/boreas/boreas_io.c
+++ b/boreas/boreas_io.c
@@ -199,22 +199,6 @@ put_finish_signal_on_queue (void *error)
 }
 
 /**
- * @brief delete key from hashtable
- *
- * @param key Key to delete from hashtable
- * @param value
- * @param hashtable   table to remove keys from
- *
- */
-static void
-exclude (gpointer key, __attribute__ ((unused)) gpointer value,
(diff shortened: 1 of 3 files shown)


boreas/boreas_io.c
```diff
@@ -30,6 +30,54 @@
...
 #define G_LOG_DOMAIN "alive scan"
 
+/**
+ * @brief Send Message about not vuln scanned alive hosts to ospd-openvas.
+ *
+ * @param num_not_scanned Number of alive hosts which were not vuln scanned.
+ * @return 0 on success, else Error.
+ */
+static int
+send_msg_about_not_vuln_scanned_hosts_to_ospd_openvas (
```
mattmundell, line 40: Normally we put the parenthesis down on the next line too in this situation.

mattmundell: But how about just using a simpler function name, like send_limit_msg?

PR #357, opened 2020-06-26
Modify osp_get_vts_version()

jjnicola:
As get_vts OSP command was extended with a new attribute version_only,
the function use get_vts with this attribute instead of the get_version command,
because now all wrappers have a vts collection and therefore, not always
get_version response has a vts entity.

### DIFF
--- a/osp/osp.c
+++ b/osp/osp.c
@@ -335,15 +335,30 @@ err_get_version:
  * @return 0 if success, 1 if error.
  */
 int
-osp_get_vts_version (osp_connection_t *connection, char **vts_version)
+osp_get_vts_version (osp_connection_t *connection, char **vts_version,
+                     char **error)
 {
-  entity_t entity, vts, version;
+  entity_t entity, vts;
+  const char *version;
+  const char *status, *status_text;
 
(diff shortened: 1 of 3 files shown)


osp/osp.c
```diff
@@ -335,34 +335,43 @@ osp_get_version (osp_connection_t *connection, char **s_name, char **s_version,
  * @return 0 if success, 1 if error.
  */
 int
-osp_get_vts_version (osp_connection_t *connection, char **vts_version)
+osp_get_vts_version (osp_connection_t *connection, char **vts_version,
+                     char **error)
```
mattmundell, line 339: Missing doc.

osp/osp.c
```diff
@@ -410,6 +419,13 @@ osp_get_vts_ext (osp_connection_t *connection, osp_get_vts_opts_t opts,
   if (vts == NULL)
     return 1;
 
+  if (opts.version_only == 1)
```
mattmundell, line 422: What was the purpose of adding this?  Should osp_get_vts_version be calling osp_get_vts_ext?

PR #362, opened 2020-06-30
Ping host only once if alive

ArnoStiefvater:
Only ping a host once if it is alive.
Hosts may still be pinged twice If the list of target hosts is short and the pinged host takes longer than the specified timeout to reply.

**Checklist**:

- [ ] Tests

### DIFF
--- a/boreas/alivedetection.c
+++ b/boreas/alivedetection.c
@@ -97,9 +97,11 @@ scan (alive_test_t alive_test)
       scanner.tcp_flag = TH_ACK;
       g_hash_table_foreach (scanner.hosts_data->targethosts, send_tcp,
                             &scanner);
+      usleep (500000);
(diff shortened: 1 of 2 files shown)


boreas/alivedetection.c
```diff
@@ -97,9 +97,11 @@ scan (alive_test_t alive_test)
       scanner.tcp_flag = TH_ACK;
       g_hash_table_foreach (scanner.hosts_data->targethosts, send_tcp,
                             &scanner);
+      usleep (500000);
```
mattmundell, line 100: In gvmd we have [gvm_usleep](https://github.com/greenbone/gvmd/blob/master/src/utils.c#L61).  It does the same, but checks for interrupts from signals.  Not sure there's anything to gain in using it, but something to consider.

PR #380, opened 2020-07-16
Smooth progress

ArnoStiefvater:
Alternate between methods of alive detection when sending pings.
This way we can achieve a a smooth progress bar for ospd-openvas without the problem of needing to correct the progress in the wrong direction if a later alive test methods detected more results than a previous one.

**Checklist**:

- [ ] Tests


ArnoStiefvater:
Closed in favor of https://github.com/greenbone/gvm-libs/pull/389

PR #382, opened 2020-07-21
Log warning if wrong PLUGIN_SET format is supplied

ArnoStiefvater:
**Checklist**:

- [ ] Tests

### DIFF
--- a/util/nvticache.c
+++ b/util/nvticache.c
@@ -171,6 +171,16 @@ nvt_feed_version ()
     }
   msg_shown = 0;
   plugin_set = g_strndup (plugin_set + 14, 12);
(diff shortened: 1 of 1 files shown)


util/nvticache.c
```diff
@@ -171,6 +171,16 @@ nvt_feed_version ()
     }
   msg_shown = 0;
   plugin_set = g_strndup (plugin_set + 14, 12);
+  if (g_strstr_len (plugin_set, -1, "\"") || g_strstr_len (plugin_set, -1, ";"))
+    {
+      g_warning ("nvt_feed_version: Erroneous %s format. Format of PLUGIN_SET "
```
mattmundell, line 176: Would be good to use __func__ in the function in the future.

PR #387, opened 2020-08-21
Separate duplicated hosts from removed hosts in the hosts structure.

jjnicola:
Also add a function to access this variable directly.


bjoernricks (approved):

mattmundell:
Thanks @bjoernricks.  Had a look last night, but don't know this code so well.

Issue #395, opened 2020-08-31
[20.8.0] unable to build the lib, during mising defines


tuxmaster5000:
OS: CentOS 7.8

Error first:
```
/builddir/build/BUILD/gvm-libs-20.8.0/boreas/alivedetection.c: In function 'scan':
/builddir/build/BUILD/gvm-libs-20.8.0/boreas/alivedetection.c:102:26: error: 'TH_SYN' undeclared (first use in this function)
       scanner.tcp_flag = TH_SYN; /* SYN */
                          ^
/builddir/build/BUILD/gvm-libs-20.8.0/boreas/alivedetection.c:102:26: note: each undeclared identifier is reported only once for each function it appears in
/builddir/build/BUILD/gvm-libs-20.8.0/boreas/alivedetection.c:110:26: error: 'TH_ACK' undeclared (first use in this function)
       scanner.tcp_flag = TH_ACK; /* ACK */
                          ^
make[2]: *** [boreas/CMakeFiles/gvm_boreas_shared.dir/alivedetection.c.o] Error 1
make[2]: Leaving directory `/builddir/build/BUILD/gvm-libs-20.8.0'
make[1]: *** [boreas/CMakeFiles/gvm_boreas_shared.dir/all] Error 2
```
Error second:
```
builddir/build/BUILD/gvm-libs-20.8.0/base/networking.c: In function 'ip_islocalhost':
/builddir/build/BUILD/gvm-libs-20.8.0/base/networking.c:841:18: error: 'addr.s_addr' may be used uninitialized in this function [-Werror=maybe-uninitialized]
               if (addr_p->s_addr == sin->sin_addr.s_addr)
                  ^
cc1: all warnings being treated as errors
make[2]: *** [base/CMakeFiles/gvm_base_shared.dir/networking.c.o] Error 1
```

marckll:
Hello @tuxmaster5000,
 
I am trying to compile gvm-libs on RHEL 7.9 and I am facing the same issue with TH_SYN and TH_ACK undeclared. I've applied your fix #396 but I am still getting the same error.

I've also tried to set directly the correct flags in the cmake command line : 
`cmake3 -DCMAKE_C_COMPILER=gcc -DCMAKE_C_FLAGS="-std=gnu99" -DCMAKE_INSTALL_PREFIX=/opt/gvm .. `
But still the same error. 

Can you confirm your fix only concerns the file boreas/CMakelists.txt with the following addition ? 
``` 
[...]

PR #396, opened 2020-08-31
Compile under CentOS 7.8

tuxmaster5000:
Fix for
https://github.com/greenbone/gvm-libs/issues/395


boreas/CMakeLists.txt
```diff
@@ -58,6 +58,13 @@ set (FILES alivedetection.c boreas_error.c boreas_io.c cli.c ping.c sniffer.c ut
 
 set (HEADERS alivedetection.h boreas_error.h boreas_io.h cli.h ping.h sniffer.h util.h)
 
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
+  if(${GLIB_VERSION} VERSION_LESS "2.50.3")
+    add_definitions(-D_BSD_SOURCE)
+  endif()
+  add_compile_options(-std=gnu99)
```
mattmundell, line 65:
```suggestion
    add_compile_options(-std=gnu99)
```

boreas/CMakeLists.txt
```diff
@@ -58,6 +58,13 @@ set (FILES alivedetection.c boreas_error.c boreas_io.c cli.c ping.c sniffer.c ut
 
 set (HEADERS alivedetection.h boreas_error.h boreas_io.h cli.h ping.h sniffer.h util.h)
 
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
+  if(${GLIB_VERSION} VERSION_LESS "2.50.3")
+    add_definitions(-D_BSD_SOURCE)
+  endif()
```
mattmundell, line 64:
```suggestion
  endif(${GLIB_VERSION} VERSION_LESS "2.50.3")
```

boreas/CMakeLists.txt
```diff
@@ -58,6 +58,13 @@ set (FILES alivedetection.c boreas_error.c boreas_io.c cli.c ping.c sniffer.c ut
...
+  add_compile_options(-std=gnu99)
+endif()
```
mattmundell, line 66:
```suggestion
endif(CMAKE_C_COMPILER_ID STREQUAL "GNU")
```

mattmundell (changes requested):
Thanks.

Do you want this in 20.08.x?  If so we should change the destination branch to greenbone:gvm-libs-20.08.

And please add a CHANGELOG.md entry.

mattmundell:
Also, would be great to have a little comment about why this is required.

tuxmaster5000:
> Also, would be great to have a little comment about why this is required.

See the the issue link, why it is needed. It looks like the code was only tested again an very new version of glibc.

bjoernricks:
> See the the issue link, why it is needed. It looks like the code was only tested again an very new version of glibc.

Our base system is Debian as you can see from our CI images. Therefore nearly all Distributions based on Debian will work without any issues. We can't invest our resources for supporting all kind of other distributions. Therefore we rely on external contributors like you to provide patches if something breaks.

tuxmaster5000:
> > See the the issue link, why it is needed. It looks like the code was only tested again an very new version of glibc.
> 
> Our base system is Debian as you can see from our CI images. Therefore nearly all Distributions based on Debian will work without any issues. We can't invest our resources for supporting all kind of other distributions. Therefore we rely on external contributors like you to provide patches if something breaks.

This is no problem,  because it is not my first patch for the greenbone world. :)

bjoernricks (approved):

Issue #397, opened 2020-08-31
[20.8.0] unable to build the lib (may be used uninitialized in this function)


tuxmaster5000:
OS: CentOS 7.8
The build fails with:
```
/builddir/build/BUILD/gvm-libs-20.8.0/boreas/util.c: In function 'set_socket':
/builddir/build/BUILD/gvm-libs-20.8.0/boreas/util.c:533:19: error: 'soc' may be used uninitialized in this function [-Werror=maybe-uninitialized]
   *scanner_socket = soc;
                   ^
cc1: all warnings being treated as errors
make[2]: *** [boreas/CMakeFiles/gvm_boreas_shared.dir/util.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
cd /builddir/build/BUILD/gvm-libs-20.8.0/util && /usr/bin/cmake3 -E cmake_symlink_library libgvm_util.so.20.8.0 libgvm_util.so.20 libgvm_util.so
make[2]: Leaving directory `/builddir/build/BUILD/gvm-libs-20.8.0'
[ 95%] Built target gvm_util_shared
make[2]: Leaving directory `/builddir/build/BUILD/gvm-libs-20.8.0'
make[1]: *** [boreas/CMakeFiles/gvm_boreas_shared.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
/builddir/build/BUILD/gvm-libs-20.8.0/base/networking.c: In function 'ip_islocalhost':
/builddir/build/BUILD/gvm-libs-20.8.0/base/networking.c:841:18: error: 'addr.s_addr' may be used uninitialized in this function [-Werror=maybe-uninitialized]
               if (addr_p->s_addr == sin->sin_addr.s_addr)
                  ^
cc1: all warnings being treated as errors
make[2]: *** [base/CMakeFiles/gvm_base_shared.dir/networking.c.o] Error 1
```

tuxmaster5000:
I think so, because as I see, the PR was merged.

PR #398, opened 2020-09-01
Fix compiler error: may be used uninitialized in this function [-Werr…

tuxmaster5000:
Fix for https://github.com/greenbone/gvm-libs/issues/397


mattmundell (changes requested):
Also for this one.  Can you make it against gvm-libs-20.08?  Otherwise it will only be released in 21.04.

tuxmaster5000:
> Also for this one. Can you make it against gvm-libs-20.08? Otherwise it will only be released in 21.04.

When I simple change the target branch then it will fail's because many other files will changed. :(

mattmundell:
> When I simple change the target branch then it will fail's because many other files will changed. :(

Yes, I had the same problem on https://github.com/greenbone/gvmd/pull/1280.

I"m not sure how the OP handled it.  I guess by using a `git rebase` command on their local tree.

bjoernricks:
`git --onto` is your fried. See https://medium.com/@gabriellamedas/git-rebase-and-git-rebase-onto-a6a3f83f9cce for some details.

ArnoStiefvater:
Thank you for contributing! It appears this fix was already applied to a different branch with https://github.com/greenbone/gvm-libs/pull/448 and merged some time later into 21.04 and master, therefore closing the PR.

PR #410, opened 2020-10-15
Add function to move a host to the end of the list

jjnicola:
**What**:
Allows to put a host in the hosts structure at the end
of the host list without reset the iterator.

**Why**:
It was need to move the host to the end without resetting the iterator. 

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/base/hosts_tests.c
+++ b/base/hosts_tests.c
@@ -212,6 +212,42 @@ Ensure (hosts, gvm_hosts_new_with_max_returns_error)
   assert_that (gvm_hosts_new_with_max ("127.0.0.1|127.0.0.2", 2), is_null);
 }
 
+Ensure (hosts, gvm_hosts_move_host_to_end)
+{
+  gvm_hosts_t *hosts = NULL;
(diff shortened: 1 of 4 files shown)


jjnicola:
Pending unittest will be done in a separated PR.

base/hosts_tests.c
```diff
@@ -212,6 +212,42 @@ Ensure (hosts, gvm_hosts_new_with_max_returns_error)
...
+
+  hosts = gvm_hosts_new ("192.168.0.0/28");
+
+  // Get first host
+  host = gvm_hosts_next (hosts);
+
+  totalhosts = gvm_hosts_count (hosts);
+  assert_that (totalhosts, is_equal_to (14));
+
+  while (g_strcmp0 (gvm_host_value_str (host), "192.168.0.9"))
```
ArnoStiefvater, line 230: Just a minor nitpick. There is a `is_equal_to_string()` function for easy use. But it does not really matter.

ArnoStiefvater (approved):

PR #414, opened 2020-10-26
Restructure LDAP / RADIUS install instructions. Fix typo.

cfi-gb:
LDAP and RADIUS auth are both optional features:

https://github.com/bjoernricks/gvm-libs/blob/b0e1f5ef9916062a2b48b84004dc2922c2ef5a08/util/CMakeLists.txt#L94-L132

so we should document / mention them separately.

This can also help with questions like https://github.com/Atomicorp/gvm/issues/36

Part of https://github.com/greenbone/templates/issues/17

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -53,7 +53,7 @@ Install prerequisites on Debian GNU/Linux 'Buster' 10:
 Prerequisites for Optional Features
 -----------------------------------
 
-Certain features of gvm-libs are optional and requires the following:
+Certain features of gvm-libs are optional and require the following:
 
 Prerequisites for LDAP authentication:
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -49,13 +45,30 @@ Install prerequisites on Debian GNU/Linux 'Buster' 10:
...
     libhiredis-dev \
     libxml2-dev \
-    libradcli-dev \
     libpcap-dev
 
 
+Prerequisites for Optional Features
+-----------------------------------
+
+Certain features of gvm-libs are optional and requires the following:
```
mattmundell, line 56: "requires" should be "require"

mattmundell (changes requested):
Thanks.

mattmundell (approved):

PR #428, opened 2021-01-26
Fix static code analysis warnings

ArnoStiefvater:
**What**:

Fix warnings of static code analysis.

**Why**:

Improve code quality and make CI happy.

**How**:

Install `clang-tools-7` and run static code analysis with e.g.:

`mkdir build && cd build/ && /bin/scan-build-7 cmake -DCMAKE_BUILD_TYPE=Debug .. && /bin/scan-build-7 -o tmp/scan-build-report make && [ -z "$(ls -A tmp/scan-build-report/)" ]`

**Checklist**:

- [ ] Tests

### DIFF
--- a/base/logging.c
+++ b/base/logging.c
@@ -480,7 +480,8 @@ gvm_log_func (const char *log_domain, GLogLevelFlags log_level,
               log_domain_entry = entry;
 
               /* Get the struct contents. */
-              prepend_format = log_domain_entry->prepend_string;
+              if (log_domain_entry->prepend_string)
+                prepend_format = log_domain_entry->prepend_string;
               time_format = log_domain_entry->prepend_time_format;
               log_file = log_domain_entry->log_file;
               if (log_domain_entry->default_level)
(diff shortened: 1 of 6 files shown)


base/networking.c
```diff
@@ -782,9 +782,9 @@ ip_islocalhost (struct sockaddr_storage *storage)
   struct in_addr addr;
   struct in_addr *addr_p;
   struct in6_addr addr6;
-  struct in6_addr *addr6_p;
+  struct in6_addr *addr6_p = NULL;
```
mattmundell, line 785: These look wrong, because they're set directly below.  Wouldn't recommend adding them just to quiet clang, but if you want to I recommend adding a note that it's for clang.

base/logging.c
```diff
@@ -510,7 +510,7 @@ gvm_log_func (const char *log_domain, GLogLevelFlags log_level,
   /* Make the tmp pointer (for iteration) point to the format string. */
   tmp = prepend_format;
 
-  while (*tmp != '\0')
```
mattmundell, line 513: I guess there should rather be a NULL check at line 483 like there is at line 441, to prevent prepend_format from ever being NULL.

mattmundell: I don't think it's good to let tmp be NULL because then prepend_buf ends up "" instead of the default "%t %s %p - ".

mattmundell (changes requested):

mattmundell (approved):

Issue #440, opened 2021-02-08
[20.8.1] Build fails uisng gcc 9.3.1


tuxmaster5000:
When I try to build the lib using gcc 9.3.1 then the build fails with:

> make[2]: Leaving directory `/builddir/build/BUILD/gvm-libs-20.8.1'
> [ 90%] Built target gvm_boreas_shared
> In file included from /usr/include/string.h:638,
>                  from /usr/include/glib-2.0/glib/gtestutils.h:30,
>                  from /usr/include/glib-2.0/glib.h:82,
>                  from /builddir/build/BUILD/gvm-libs-20.8.1/base/settings.h:30,
>                  from /builddir/build/BUILD/gvm-libs-20.8.1/base/prefs.c:28:
> In function 'strncpy',
>     inlined from 'prefs_config' at /builddir/build/BUILD/gvm-libs-20.8.1/base/prefs.c:155:3:
> /usr/include/bits/string3.h:120:10: error: '__builtin_strncpy' specified bound 2048 equals destination size [-Werror=stringop-truncation]
>   120 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> cc1: all warnings being treated as errors
> make[2]: *** [base/CMakeFiles/gvm_base_shared.dir/prefs.c.o] Error 1
> make[2]: Leaving directory `/builddir/build/BUILD/gvm-libs-20.8.1'
> make[1]: *** [base/CMakeFiles/gvm_base_shared.dir/all] Error 2
> make[1]: *** Waiting for unfinished jobs....
> In file included from /usr/include/string.h:638,
>                  from /usr/include/glib-2.0/glib/gtestutils.h:30,
>                  from /usr/include/glib-2.0/glib.h:82,
>                  from /builddir/build/BUILD/gvm-libs-20.8.1/util/../base/nvti.h:30,
>                  from /builddir/build/BUILD/gvm-libs-20.8.1/util/kb.h:28,
>                  from /builddir/build/BUILD/gvm-libs-20.8.1/util/kb.c:27:
> In function 'strncpy',
>     inlined from 'redis_direct_conn' at /builddir/build/BUILD/gvm-libs-20.8.1/util/kb.c:429:3:
[...]

PR #441, opened 2021-02-08
Fix for compile it under gcc 9.3.1

tuxmaster5000:
Fix for https://github.com/greenbone/gvm-libs/issues/440

### DIFF
--- a/base/prefs.c
+++ b/base/prefs.c
@@ -166,10 +166,10 @@ prefs_config (const char *config)
   if (!global_prefs)
     prefs_init ();
 
-  #pragma GCC diagnostic push
-  #pragma GCC diagnostic ignored "-Wstringop-truncation"
-    strncpy (buffer, config, sizeof (buffer));
-  #pragma GCC diagnostic pop
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+  strncpy (buffer, config, sizeof (buffer));
+#pragma GCC diagnostic pop
   if (!init_settings_iterator_from_file (&settings, buffer, "Misc"))
     {
       while (settings_iterator_next (&settings))
--- a/base/prefs.c
+++ b/base/prefs.c
@@ -166,7 +166,10 @@ prefs_config (const char *config)
   if (!global_prefs)
     prefs_init ();
 
-  strncpy (buffer, config, sizeof (buffer));
+  #pragma GCC diagnostic push
+  #pragma GCC diagnostic ignored "-Wstringop-truncation"
+    strncpy (buffer, config, sizeof (buffer));
+  #pragma GCC diagnostic pop
   if (!init_settings_iterator_from_file (&settings, buffer, "Misc"))
     {
       while (settings_iterator_next (&settings))
--- a/base/prefs.c
+++ b/base/prefs.c
@@ -166,10 +166,7 @@ prefs_config (const char *config)
   if (!global_prefs)
     prefs_init ();
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstringop-truncation"
-  strncpy (buffer, config, sizeof (buffer));
-#pragma GCC diagnostic pop
+  strncpy (buffer, config, sizeof (buffer) - 1);
   if (!init_settings_iterator_from_file (&settings, buffer, "Misc"))
     {
       while (settings_iterator_next (&settings))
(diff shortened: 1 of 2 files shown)


util/kb.c
```diff
@@ -393,7 +393,11 @@ redis_new (kb_t *kb, const char *kb_path)
 
   kbr = g_malloc0 (sizeof (struct kb_redis) + strlen (kb_path) + 1);
   kbr->kb.kb_ops = &KBRedisOperations;
-  strncpy (kbr->path, kb_path, strlen (kb_path));
+  #pragma GCC diagnostic push
+  #pragma GCC diagnostic ignored "-Wstringop-truncation"
+  #pragma GCC diagnostic ignored "-Wstringop-overflow"
+    strncpy (kbr->path, kb_path, strlen (kb_path));
```
y0urself, line 399: I guess the more correct way would be fixing it, instead of ingoring.
Afaik the problem is, that `strncpy` uses the last byte to null-terminate the copy  with `\0`.
So the compiler warns, that if the string is really as long as `kb_path` it can't be null-terminated.
The compiler assumes that the given string is not terminated.
I guess we should do something like:
```suggestion
    strncpy (kbr->path, kb_path, strlen (kb_path) - 1);
```

tuxmaster5000: I have updated the PR, so the stringop-truncation statement can be removed,

base/prefs.c
```diff
@@ -166,7 +166,10 @@ prefs_config (const char *config)
   if (!global_prefs)
     prefs_init ();
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
   strncpy (buffer, config, sizeof (buffer));
+#pragma GCC diagnostic pop
```
y0urself, line 172:
```suggestion
  strncpy (buffer, config, sizeof (buffer) - 1);
```

util/kb.c
```diff
@@ -393,7 +393,10 @@ redis_new (kb_t *kb, const char *kb_path)
 
   kbr = g_malloc0 (sizeof (struct kb_redis) + strlen (kb_path) + 1);
   kbr->kb.kb_ops = &KBRedisOperations;
-  strncpy (kbr->path, kb_path, strlen (kb_path));
+  #pragma GCC diagnostic push
+  #pragma GCC diagnostic ignored "-Wstringop-overflow"
+    strncpy (kbr->path, kb_path, strlen (kb_path) - 1);
+  #pragma GCC diagnostic pop
```
y0urself, line 399:
```suggestion
    strncpy (kbr->path, kb_path, strlen (kb_path) - 1);
```

y0urself (commented):
So we would need to change every occurence ... but I am not sure if it works this way. I will look into it later.

util/kb.c
tuxmaster5000: This change will not work:
[ 97%] Built target gvm_base_shared
In file included from /usr/include/string.h:638,
                 from /usr/include/glib-2.0/glib/gtestutils.h:30,
                 from /usr/include/glib-2.0/glib.h:82,
                 from /builddir/build/BUILD/gvm-libs-20.8.1/util/../base/nvti.h:30,
                 from /builddir/build/BUILD/gvm-libs-20.8.1/util/kb.h:28,
                 from /builddir/build/BUILD/gvm-libs-20.8.1/util/kb.c:27:
In function 'strncpy',
    inlined from 'redis_new' at /builddir/build/BUILD/gvm-libs-20.8.1/util/kb.c:396:3:
/usr/include/bits/string3.h:120:10: error: '__builtin_strncpy' specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
  120 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/builddir/build/BUILD/gvm-libs-20.8.1/util/kb.c: In function 'redis_new':
/builddir/build/BUILD/gvm-libs-20.8.1/util/kb.c:396:32: note: length computed here
  396 |   strncpy (kbr->path, kb_path, strlen (kb_path) - 1);
      |                                ^~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

jjnicola:
Hi @tuxmaster5000,

thank you very much for this PR. Unfortunately, some extra work was necessary to fix this compiler warning which ended in a code improvement.
Therefore, I am closing this PR in favor of #443
Best Regards.

PR #443, opened 2021-02-10
Code improvement. Also fix compiler warning.

jjnicola:
**What**:
Use a char pointer instead of an zero-lenght array as kb_redis struct member.
Use g_strdup instead of strncpy for initializing kbr->path.
Free the path where necessary.

**Why**:
> Declaring zero-length arrays in other contexts, including as interior members of structure objects or as non-member objects, is discouraged. Accessing elements of zero-length arrays declared in such contexts is undefined and may be diagnosed.

https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 ### Changed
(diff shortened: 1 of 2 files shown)


jjnicola:
Close #440

bjoernricks (approved):

PR #464, opened 2021-03-19
Fix finish_signal_on_queue

jjnicola:
**What**:
Fix finish_signal_on_queue. Check for the signal in the whole items under the key, instead of
just checking the the last item.

**Why**:
In some cases the `finished` was not reallocated causing interrupted scans

**How**:
Run a scan with Boreas alive test, with `expand_vhosts` option enabled and `allow_simultaneous_ips` disabled. Check that the scan is not interrupted and all hosts are scanned.

**Checklist**:

- [ ] Tests
- [X] [CHANGELOG](https://github.com/greenbone/gvm-libs/blob/master/CHANGELOG.md) Entry

### DIFF
--- a/boreas/boreas_io.c
+++ b/boreas/boreas_io.c
@@ -240,7 +240,7 @@ put_host_on_queue (kb_t kb, char *addr_str)
 }
 
 /**
- * @brief Checks if the finish signal is already set as last item in the queue.
+ * @brief Checks if the finish signal is already set.
  *
  * @param main_kb  kb to use
  * @return 1 if it is already set. 0 otherwise.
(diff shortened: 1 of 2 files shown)


boreas/boreas_io.c
```diff
@@ -240,26 +240,38 @@ put_host_on_queue (kb_t kb, char *addr_str)
...
 
-  if (last_queue_item && (last_queue_item->type == KB_TYPE_STR)
-      && (!g_strcmp0 (last_queue_item->v_str, ALIVE_DETECTION_FINISHED)))
-    ret = 1;
+  if (fin_msg_already_on_queue)
+    return 1;
 
-  kb_item_free (last_queue_item);
+  /* Check if it was already set throught the whole items under the key.
+     If so, set the static variable to avoid quering redis unnecessarily. */
```
ArnoStiefvater, line 259:
```suggestion
     If so, set the static variable to avoid querying redis unnecessarily. */
```

ArnoStiefvater (approved):

Issue #476, opened 2021-04-16
[21.4.0] build fails with gcc error


tuxmaster5000:
The build fails with:
/builddir/build/BUILD/gvm-libs-21.4.0/boreas/alivedetection.c: In function 'start_alive_detection':
/builddir/build/BUILD/gvm-libs-21.4.0/boreas/alivedetection.c:271:64: error: 'prev_alive' may be used uninitialized in this function [-Werror=maybe-uninitialized]
           number_of_dead_hosts = remaining_batch - (curr_alive - prev_alive);
                                                    ~~~~~~~~~~~~^~~~~~~~~~~~~
/builddir/build/BUILD/gvm-libs-21.4.0/boreas/alivedetection.c:82:7: note: 'prev_alive' was declared here
   int prev_alive;
       ^~~~~~~~~~
/builddir/build/BUILD/gvm-libs-21.4.0/boreas/alivedetection.c:271:32: error: 'remaining_batch' may be used uninitialized in this function [-Werror=maybe-uninitialized]
           number_of_dead_hosts = remaining_batch - (curr_alive - prev_alive);
           ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/builddir/build/BUILD/gvm-libs-21.4.0/boreas/alivedetection.c:81:7: note: 'remaining_batch' was declared here
   int remaining_batch;
       ^~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[2]: *** [boreas/CMakeFiles/gvm_boreas_shared.dir/build.make:66: boreas/CMakeFiles/gvm_boreas_shared.dir/alivedetection.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Entering directory '/builddir/build/BUILD/gvm-libs-21.4.0'

y0urself:
@tuxmaster5000 can this issue be closed?

tuxmaster5000:
Yes, because the PR was merged.

PR #477, opened 2021-04-16
Fix gcc errors

tuxmaster5000:
Fix for
https://github.com/greenbone/gvm-libs/issues/476


jjnicola (dismissed):

jjnicola (dismissed):
Thanks for the fix. Could you change please the target branch to gvm-libs-21.04 instead of master ?

y0urself (approved):

PR #484, opened 2021-04-26
Introducing passwordbasedauthentication based on lcrypt and sha512

nichtsfrei:
Introducing passwordbasedauthentication to handle upgrade from MD5 to
sha512 within gvmd.

The decision to write new implementation based on lcrypt and sha512
rather than updating authutils is based on the decision to follow phc
comliance as closely as feasible to allow easier versioning and updating
of hashes for password based logins.

To allow easy handling of correct but dated hashes a new state:
RECOMMEND_UPDATE
is introduced for passwords ontop of valid/invalid.

Therefore a direct dependency of passwordbasedauthentication.c to
authutils.c is chosen to have easier access to the deprecated
functionality.

To create the hash crypt_r is chosen to have a thread safe hashing
possibility, this comes with the costs of portability since not all
crypt libraries do support that. However libcrypt (used by
debian:testing) as well as UFC crypt (used by buster) do.

Due to the limitation of salt to 16 bytes the newly introduced pepper
usage is limited to 4 bytes. This needs to be respected in using
clients, e.g. gvmd and enforced there.

**What**:

**Why**:

**How**:

**Checklist**:

- [x] Tests


util/passwordbasedauthentication.c
```diff
@@ -0,0 +1,271 @@
...
+        tmp[0] = setting->pepper[i];
+      tmp--;
+    }
+
+  data = calloc (1, sizeof (struct crypt_data));
+  rslt = crypt_r (password, settings, data);
+  if (rslt == NULL)
+    goto exit;
+  result = malloc (CRYPT_OUTPUT_SIZE);
+  memcpy (result, rslt, CRYPT_OUTPUT_SIZE);
```
bjoernricks, line 200: As discussed I am not sure if this might crash if crypt_r returns a smaller string here under some situations.

util/passwordbasedauthentication.c
```diff
@@ -0,0 +1,271 @@
...
+#ifndef CRYPT_GENSALT_OUTPUT_SIZE
+#define CRYPT_GENSALT_OUTPUT_SIZE 192
+#endif
+
+#ifndef CRYPT_OUTPUT_SIZE
+#define CRYPT_OUTPUT_SIZE 384
+#endif
+
+int
+is_prefix_not_supported (const char *id)
```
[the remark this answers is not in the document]
nichtsfrei: you're right; changed to is_prefix_supported.

util/passwordbasedauthentication.c
nichtsfrei: changed to MIN(strlen(rslt), CRYPT_OUTPUT_SIZE)

util/passwordbasedauthentication.c
```diff
@@ -94,7 +96,7 @@ crypt_gensalt_r (const char *prefix, unsigned long count, const char *rbytes,
   unsigned int written = 0, used = 0;
   unsigned long value = 0;
   if ((rbytes != NULL && nrbytes < 3) || output_size < 16
-      || is_prefix_not_supported (prefix))
+      || is_prefix_supported (prefix) == 0)
```
bjoernricks, line 99:
```suggestion
      || !is_prefix_supported (prefix))
```

util/passwordbasedauthentication.c
```diff
@@ -142,7 +144,7 @@ pba_init (const char *pepper, unsigned int pepper_size, unsigned int count,
   struct PBASettings *result = NULL;
   if (pepper_size > MAX_PEPPER_SIZE)
     goto exit;
-  if (prefix != NULL && is_prefix_not_supported (prefix))
+  if (prefix != NULL && is_prefix_supported (prefix) == 0)
```
bjoernricks, line 147:
```suggestion
  if (prefix != NULL && !is_prefix_supported (prefix))
```

util/passwordbasedauthentication.c
```diff
@@ -178,7 +180,7 @@ pba_hash (struct PBASettings *setting, const char *password)
 
   if (!setting || !password)
     goto exit;
-  if (is_prefix_not_supported (setting->prefix) != 0)
+  if (is_prefix_supported (setting->prefix) == 0)
```
bjoernricks, line 183:
```suggestion
  if (!is_prefix_supported (setting->prefix))
```

bjoernricks (approved):

PR #502, opened 2021-05-18
Integrate sentry

jjnicola:
**What**:
Integrate sentry

**Why**:

**How**:
Set the environment variable BUILD_SENTRY=1 before running cmake.
`BUILD_SENTRY=1 cmake -DCMAKE_INSTALL_PREFIX=$INSTALL_PREFIX .. `

sentry must be installed in the same install prefix, so cmake can use `findpackage` for sentry

**Checklist**:

- [ ] Tests


base/logging.c
```diff
@@ -626,6 +628,9 @@ gvm_log_func (const char *log_domain, GLogLevelFlags log_level,
                             message);
   g_free (prepend);
 
+  if (log_level == G_LOG_LEVEL_WARNING)
```
timopollmeier, line 631: I think this should include the more critical message types (which are represented by lower numbers): 
```suggestion
  if (log_level <= G_LOG_LEVEL_WARNING)
```

base/gvm_sentry.c
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * The function does nothing if HAVE_SENTRY is not defined
+ *
+ */
+void
+gvm_close_sentry (void)
+{
+#ifdef HAVE_SENTRY
+  if (is_sentry_initialized ())
+    sentry_close ();
```
timopollmeier, line 116: Maybe this should also reset global_init_sentry back to 0.

timopollmeier (changes requested):

base/CMakeLists.txt
```diff
@@ -27,15 +27,25 @@ endif (NOT PKG_CONFIG_FOUND)
...
 
+if ($ENV{BUILD_SENTRY})
+  find_package (SENTRY REQUIRED)
+  if (${SENTRY_FOUND})
+    message ("Bulding with Sentry integration")
+    set (SENTRY_LDFLAGS "-lsentry")
+    set (SENTRY_CFLAGS "-I${CMAKE_INSTALL_PREFIX}/include")
+    add_definitions (-DHAVE_SENTRY=1)
+  endif (${SENTRY_FOUND})
+endif ($ENV{BUILD_SENTRY} AND ${SENTRY_FOUND})
```
timopollmeier, line 38:
```suggestion
endif ($ENV{BUILD_SENTRY})
```

timopollmeier (approved):

PR #519, opened 2021-06-15
Replace localtime with localtime_r

y0urself:
* Replace `localtime` with `localtime_r()`
* Also added logger warning, if `localtime_r` fails

**What**:

**Why**:

* `localtime()` has potential security leaks according to CodeQL

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/base/logging.c
+++ b/base/logging.c
@@ -82,15 +82,15 @@ gchar *
 get_time (gchar *time_fmt)
 {
   time_t now;
-  struct tm *ts;
+  struct tm ts;
   gchar buf[80];
 
   /* Get the current time. */
   now = time (NULL);
 
   /* Format and print the time, "ddd yyyy-mm-dd hh:mm:ss zzz." */
-  ts = localtime (&now);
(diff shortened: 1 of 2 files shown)


base/logging.c
```diff
@@ -82,15 +82,18 @@ gchar *
...
 
   /* Get the current time. */
   now = time (NULL);
 
   /* Format and print the time, "ddd yyyy-mm-dd hh:mm:ss zzz." */
-  ts = localtime (&now);
-  strftime (buf, sizeof (buf), time_fmt, ts);
+  if (localtime_r (&now, &ts) == NULL)
+    {
+      g_warning ("%s: localtime failed", __func__);
```
timopollmeier, line 94: I'm not sure if we should log a warning here because `get_time` is used in the log function, so this may result in an infinite recursion if the problem with the `localtime_r` call persists.

timopollmeier (changes requested):

timopollmeier (approved):

PR #520, opened 2021-06-17
Initialize reallocated memory in gvm_hosts_add

timopollmeier:
**What**:
If the array memory for the hosts is reallocated, the new memory is
initialized with all \0 bytes.

**Why**:
This ensures the array is correctly terminated when gvm_hosts_fill_gaps
is called when excluding hosts.

**How**:
By creating multiple targets with very large host ranges (> 2 million hosts) and adding excluded hosts to at least one of them.

**Checklist**:

- [ ] Tests

### DIFF
--- a/base/hosts.c
+++ b/base/hosts.c
@@ -957,7 +957,7 @@ gvm_hosts_add (gvm_hosts_t *hosts, gvm_host_t *host)
       hosts->max_size *= 4;
       hosts->hosts =
         g_realloc_n (hosts->hosts, hosts->max_size, sizeof (*hosts->hosts));
-      memset (hosts->hosts + (hosts->count * sizeof (gvm_host_t*)), '\0',
+      memset (hosts->hosts + hosts->count, '\0',
               (hosts->max_size - hosts->count) * sizeof (gvm_host_t*));
     }
   hosts->hosts[hosts->count] = host;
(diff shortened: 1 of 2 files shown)


base/hosts.c
```diff
@@ -957,6 +957,8 @@ gvm_hosts_add (gvm_hosts_t *hosts, gvm_host_t *host)
       hosts->max_size *= 4;
       hosts->hosts =
         g_realloc_n (hosts->hosts, hosts->max_size, sizeof (*hosts->hosts));
+      memset (hosts->hosts + (hosts->count * sizeof (gvm_host_t*)), '\0',
+              (hosts->max_size - hosts->count) * sizeof (gvm_host_t*));
```
y0urself, line 961:
```suggestion
              (hosts->max_size - hosts->count) * sizeof (gvm_host_t *));
```

base/hosts.c
```diff
@@ -957,6 +957,8 @@ gvm_hosts_add (gvm_hosts_t *hosts, gvm_host_t *host)
       hosts->max_size *= 4;
       hosts->hosts =
         g_realloc_n (hosts->hosts, hosts->max_size, sizeof (*hosts->hosts));
+      memset (hosts->hosts + (hosts->count * sizeof (gvm_host_t*)), '\0',
```
y0urself, line 960:
```suggestion
      memset (hosts->hosts + (hosts->count * sizeof (gvm_host_t *)), '\0',
```

y0urself (changes requested):

y0urself (approved):

PR #559, opened 2021-07-07
[21.04] Fix and integrate the new Dockerimage stucture into 21.04

y0urself:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests


.docker/testing/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+    make \
+    pkg-config \
+    uuid-dev \
+    libssl-dev \
+    lcov \
+    libical-dev \
+    libpq-dev \
+    libnet1-dev \
+    postgresql-server-dev-all \
+    xsltproc \
```
bjoernricks, line 44: i am not sure about the dependencies. looks like the dependencies from gvmd to me. i don't think gvm-libs needs postgres at all and also not xsltproc. but maybe i am wrong.

.docker/testing/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+# & install cgreen for unit tests
+RUN git clone --depth 1 https://github.com/cgreen-devs/cgreen.git \
+    && cd cgreen \
+    && make \
+    && make test \
+    && make install \
+    && cd .. \
+    && rm -rf cgreen
+
+ENV LD_LIBRARY_PATH="/usr/local/lib"
```
bjoernricks, line 81: IMHO this should not be necessary but it doesn't harm.

y0urself: afaik it did not work without it, but I can try to remove it and run the CI again without it ...

bjoernricks (approved):

y0urself:
addressing AP-1493

PR #561, opened 2021-07-14
Do not send unnecessary message

ArnoStiefvater:
**What**:

Do not send unnecessary message anymore.
When 0 alive hosts are left to scan and max_scan_hosts limit is reached no message will be generated for that case anymore.

**Why**:

SC-298

**How**:

Start a scan with 1 host and `max_scan_hosts = 1` and `test_alive_hosts_only = yes`. Before PR a Error Message message about 0 left alive hosts is generated. After PR no message in generated anymore.

**Checklist**:

- [ ] Tests

### DIFF
--- a/boreas/boreas_io.c
+++ b/boreas/boreas_io.c
@@ -185,7 +185,10 @@ get_host_from_queue (kb_t alive_hosts_kb, gboolean *alive_deteciton_finished)
 
               num_not_scanned_hosts =
                 get_alive_hosts_count () - get_max_scan_hosts ();
-              send_limit_msg (num_not_scanned_hosts);
+              if (0 != num_not_scanned_hosts)
+                {
+                  send_limit_msg (num_not_scanned_hosts);
+                }
             }
           g_debug ("%s: Boreas already finished scanning and we reached the "
                    "end of the Queue of alive hosts.",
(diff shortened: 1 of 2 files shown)


ArnoStiefvater:
SC-298

boreas/boreas_io.c
```diff
@@ -185,7 +185,10 @@ get_host_from_queue (kb_t alive_hosts_kb, gboolean *alive_deteciton_finished)
 
               num_not_scanned_hosts =
                 get_alive_hosts_count () - get_max_scan_hosts ();
-              send_limit_msg (num_not_scanned_hosts);
+              if (0 != num_not_scanned_hosts)
```
paulnaumann, line 188: I would prefer to do the check like this              
`if (num_not_scanned_hosts > 0)` 
because that way it is not dependent on max_scan_host_reached()

paulnaumann (approved):

y0urself:
SC298

boreas/boreas_io.c
ArnoStiefvater: But the message only has to be sent when `max_scan_host_reached()` returns true. So its dependent on it on purpose.

jjnicola (approved):

PR #568, opened 2021-07-26
Get vector

jjnicola:
**What**:
Add function to get the severity_vector, otherwise the cvss_base_vector.

**Why**:
Extended severity was introduced and there are new tags to hold severity information. severity_vector is the tag name.
Some VTs doesn't have the new tag, but the old one cvss_base_vector. This function returns the vector severity_vector, or the cvss_base_vector if the vt does not have the new tag.

**How**:
Test added

**Checklist**:

- [x] Tests

### DIFF
--- a/base/nvti.c
+++ b/base/nvti.c
@@ -937,6 +937,36 @@ nvti_severity_score (const nvti_t *n)
 }
 
 /**
+ * @brief Get the severity score
+ *
+ * Extended severity was introduced but still not all
+ * vts are using it. Therefore it must be check if
+ * we can calculate the score from the severity_vector tag
+ * or we have to calculate it from the deprecated
+ * cvss_base_vector tag.
+ *
+ * @param n The NVT Info structure.
+ *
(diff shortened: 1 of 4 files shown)


base/nvti.c
```diff
@@ -936,6 +936,36 @@ nvti_severity_score (const nvti_t *n)
   return score;
 }
 
+/**
+ * @brief Get the severity score
+ *
+ * Extended severity was introduced but still not all
+ * vts are using it. Therefore it must be check if
```
ArnoStiefvater, line 943:
```suggestion
 * vts are using it. Therefore it must be checked if
```

base/nvti.c
```diff
@@ -936,6 +936,36 @@ nvti_severity_score (const nvti_t *n)
...
+ * we can calculate the score from the severity_vector tag
+ * or we have to calculate it from the deprecated
```
ArnoStiefvater, line 945:
```suggestion
 * or if we have to calculate it from the deprecated
```

ArnoStiefvater (approved):

PR #574, opened 2021-08-04
Use better defaults for installation directories

bjoernricks:
**What**:

Changed defaults for installation locations

* SYSCONFDIR is /etc by default now
* LOCALSTATEDIR is /var by default now
* Introduced GVM_RUN_DIR that uses /run/gvm by default
* GVM_PID_DIR is /run/gvm  by default now

**Why**:

Use same defaults in every component and use defaults that make sense on a normal system.

**How**:

**Checklist**:

- [ ] Tests


y0urself (approved):

wisukind:
Seems like this patch broke useage of GVM_INSTALL_PREFIX variable. Whatever you put in that variable, this is now completely ignored by cmake at compile time, and the new defaults are systematically used.

bjoernricks:
That's not correct. We just follow another scheme. CMAKE_INSTALL_PREFIX is still used for the application data (the files that are normally installed somewhere into /usr). state or config data is installed into root directories by default for example into /etc, /var or /run because it doesn't really make sense to install them into a different location by default.

The aim of the new install locations is to provide and easy system wide installation by default.

bjoernricks:
If you don't use a system wide installation you need to adjust the cmake parameters listed above in the PR description.

wisukind:
I understand the logic but why ignoring GVM_INSTALL_PREFIX ?  Wouldn't it be more simple to provide those defaults in case GVM_INSTALL_PREFIX is unset, which would apply to most users not wanting a particular root installation path by default without breaking those who wants ? Now you're forcing users to use this new default, unless they change every single state or config data manually.

bjoernricks:
Just to be clear there is no GVM_INSTALL_PREFIX it is CMAKE_INSTALL_PREFIX. CMAKE_INSTALL_PREFIX is never empty. It's default is /usr/local. Therefore there is no possibility to check if it is set or not.

LZD-TMoreggia:
So if we used CMAKE_INSTALL_PREFIX before and we want to keep using same structure, now we need to manually set

- SYSCONFDIR to ${CMAKE_INSTALL_PREFIX}/etc
- LOCALSTATEDIR to ${CMAKE_INSTALL_PREFIX}/var
- GVM_RUN_DIR to ${LOCALSTATEDIR}/run

is that correct?

ArnoStiefvater:
There is no GVM_PID_DIR in gvm-libs anymore. 

And I have GVM_RUN_DIR set to `<install_prefix>/run/gvm` as the new default is `run/gvm`. Everything else seems to be correct.

PR #577, opened 2021-08-04
Add: subscribe and retrieve message from mqtt

nichtsfrei:
To retrieve messages

1. mqtt_subscribe
1. mqtt_retrieve_message
1. mqtt_unsubscribe

are implemented.

Since at least in openvas most retrieving functionalities are done a in
request/response fashion `mqtt_retrieve_message` is written in a
blocking way instead of using the callback functionality of paho.

With that we have fine granular control over threads and processes and
can decide how we want to treat message retrieval in more abstract
libraries.

There is a small example on how to use the mqtt_* api retrieve
messages:

```
#include <gvm/util/mqtt.h>
#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#define TOPIC "#"

void
signalhandler (int signum)
{
  printf ("Caught %d, reset mqtt\n", signum);
  mqtt_unsubscribe (TOPIC);
  mqtt_reset ();
  exit (0);
}

int
main ()
{
  signal (SIGINT, signalhandler);

  int rc;
  if ((rc = mqtt_init ("localhost:9138")) != 0)
    {
      printf ("init returned %d, quitting\n", rc);
      goto exit;
    }

  if ((rc = mqtt_subscribe (TOPIC)) != 0)
    {
      printf ("subscribe returned %d, quitting\n", rc);
      goto exit;
    }

  while (1)
    {
      char *topic;
      int topic_len;
      char *payload;
      int payload_len;
      if ((rc =
             mqtt_retrieve_message (&topic, &topic_len, &payload, &payload_len))
          == -1)
        {
          printf ("unable to retrieve message, quitting\n");
          goto exit;
        }
      if (rc == 0)
        printf ("[%d] Got message: %s on topic %s\n", payload_len,
                payload != NULL ? payload : "N/A",
                topic != NULL ? topic : "N/A");
      else
        rc = 0;
      if (payload != NULL)
        free (payload);
      if (topic != NULL)
        free (topic);
    }
exit:
  mqtt_unsubscribe (TOPIC);
  mqtt_reset ();
  return rc;
}
```

to compile it please install libgvm and build it via:

```
gcc $(pkg-config --cflags glib-2.0) main.c -lgvm_util -lgvm_base $(pkg-config --libs glib-2.0)
```


util/mqtt.c
```diff
@@ -491,3 +489,235 @@ mqtt_publish_single_message (const char *server_uri_in, const char *topic,
...
+ * failed.
+ */
+int
+mqtt_subscribe_r (mqtt_t *mqtt, int qos, const char *topic)
+{
+  int rc = 0;
+  if (mqtt == NULL || mqtt->client == NULL)
+    {
+      rc = -1;
+      goto exit;
```
y0urself, line 516:
```suggestion
      return rc;
```

util/mqtt.c
```diff
@@ -491,3 +489,235 @@ mqtt_publish_single_message (const char *server_uri_in, const char *topic,
...
+    }
+  MQTTSubscribe_options opts = MQTTSubscribe_options_initializer;
+  MQTTProperties props = MQTTProperties_initializer;
+  MQTTResponse resp =
+    MQTTClient_subscribe5 (mqtt->client, topic, qos, &opts, &props);
+  if (resp.reasonCode != MQTTREASONCODE_GRANTED_QOS_1)
+    {
+      rc = -2;
+    }
+exit:
```
y0urself, line 526:
```suggestion
```
You only use the `goto` once ... I am no fan of `goto` because it jumps in the code and makes it unreadable ... also it is only used once, so I do not see any advantage of it here ...

util/mqtt.c
```diff
@@ -491,3 +489,235 @@ mqtt_publish_single_message (const char *server_uri_in, const char *topic,
...
+ * failed.
+ */
+int
+mqtt_unsubscribe_r (mqtt_t *mqtt, const char *topic)
+{
+  int rc = 0;
+  if (mqtt == NULL || mqtt->client == NULL)
+    {
+      rc = -1;
+      goto exit;
```
y0urself, line 569:
```suggestion
      return rc;
```

util/mqtt.c
```diff
@@ -491,3 +489,235 @@ mqtt_publish_single_message (const char *server_uri_in, const char *topic,
...
+    }
+
+  if ((rc = MQTTClient_unsubscribe (mqtt->client, topic)) != MQTTCLIENT_SUCCESS)
+    {
+      rc = -2;
+      goto exit;
```
y0urself, line 575:
```suggestion
```

util/mqtt.c
```diff
@@ -491,3 +489,235 @@ mqtt_publish_single_message (const char *server_uri_in, const char *topic,
...
+    }
+
+exit:
```
y0urself, line 578:
```suggestion
```

.gitignore
```diff
@@ -1,2 +1,9 @@
 build/
 .ccls
+# clangd
```
ArnoStiefvater, line 3: Is there a good reason for including those? I don't really mind having them. just want to know because
CMakeFiles and compile_commands.json are already included in the build folder for me.

util/mqtt.c
```diff
@@ -38,7 +38,11 @@
  * is established every call.
  */
 
+#include <stdlib.h>
+#include <string.h>
+#define _GVM_MQTT_INTERNAL
```
ArnoStiefvater, line 43: Why is this needed together with moving the typedef to the header?
I can compile and run the example without these changes. 
mqtt_connect() can stay static too.

.gitignore
nichtsfrei: I don't work in the build dir hence I put the extra files into .gitignore to not commit them; what's the issue?

util/mqtt.c
nichtsfrei: That was done before I switched back to the global setup and therefore was using mqtt_t outside of mqtt as void but wanted to define it once without having to cast based on the function definition. This is obsolete now and can be moved back to mqtt.c

util/mqtt.c
y0urself: This has not been changed :trollface:

util/mqtt.c
```diff
@@ -211,7 +214,7 @@ mqtt_reset ()
  * @return MQTTClient or NULL on error.
  */
 static MQTTClient
-mqtt_create (mqtt_t *mqtt)
+mqtt_create (mqtt_t *mqtt, const char *address)
```
Kraemii, line 217: Adjust doxygen

y0urself (approved):

PR #612, opened 2021-10-29
Add: automatic reinit mqtt after reset

Kraemii:
**What**:
After a fork the mqtt connection has to be reset. After that a reinitialization has to be done. Unfortunately the only way to do this was by publishing a message. I also added this reinitialization also to the subscription function.

**Why**:
In order to subscribe to a topic (after a fork) this was necessary. Else the subscription fails.

**How**:

**Checklist**:

- [ ] Tests
- [x] PR merge commit message adjusted


util/mqtt.c
```diff
@@ -355,6 +355,28 @@ mqtt_init (const char *server_uri)
   return 0;
 }
 
+/**
+ * @brief Reinitializes communication after mqtt_reset was used
+ *
+ */
+void
+mqtt_reinit ()
```
ArnoStiefvater, line 363: I would not do it in a separate function for this case. Else I think its getting a bit confusing.
We already have mqtt_reset and mqtt_init.

ArnoStiefvater (changes requested):

util/mqtt.h
```diff
@@ -47,9 +47,15 @@ int
...
 
 int
-mqtt_retrieve_message (char **, int *, char **, int *);
+mqtt_retrieve_message (char **, int *, char **, int *, const unsigned int);
 
 int
 mqtt_unsubscribe (const char *);
 
+const char *
+mqtt_reason_code_to_string (int code);
```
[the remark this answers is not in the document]
Kraemii: Added it for debugging and forgot to delete them in the mqtt.h

util/mqtt.h
```diff
@@ -47,9 +47,15 @@ int
...
+
+const char *
+mqtt_str_error (int err);
```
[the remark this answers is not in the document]
Kraemii: Added it for debugging and forgot to delete them in the mqtt.h

util/mqtt.c
Kraemii: I would rather rename it then, but as it is indeed a re-initialization for the current fork after a mqtt_reset was called. Also it is documented as that

ArnoStiefvater: I don't like that function because it does not really reinit. It first checks if it needs to reinit then reinits or does nothing.
So in 99% of cases this functions just does nothing.
If we really want to have a separate function I would do the `if ((mqtt_get_global_client ()) == NULL)` outside of this function.

util/mqtt.c
```diff
@@ -355,6 +355,26 @@ mqtt_init (const char *server_uri)
...
+{
+  const char *server_uri;
+  // init new global client if it was mqtt_reset()
```
ArnoStiefvater, line 366: Sorry for being annoying but either remove this comment or move it out of the function.

ArnoStiefvater (changes requested):

util/mqtt.c
Kraemii: It is not annoying at all. It is important to keep the code clean and I am glad that you noticed it!

ArnoStiefvater (approved):

PR #616, opened 2021-11-05
Install libcgreen unit test library via Debian

bjoernricks:
**What**:

With using Debian stable (Bullseye) a package of libcgreen is available
and it isn't required to build it ourselfs.

**Why**:

We use Debian Bullseye for the container images which provides a package for libcgreen.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/.docker/build.Dockerfile
+++ b/.docker/build.Dockerfile
@@ -36,6 +36,7 @@ RUN apt-get update && \
     libldap2-dev \
     libradcli-dev \
     libpaho-mqtt-dev \
+    libcgreen1-dev \
     && rm -rf /var/lib/apt/lists/*
 
 # Install gcc/g++ compiler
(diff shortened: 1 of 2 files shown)


bjoernricks:
The builds are failing because the images only get updated *after* the PR is merged.

y0urself (dismissed):

.docker/build.Dockerfile
```diff
@@ -36,6 +36,7 @@ RUN apt-get update && \
     libldap2-dev \
     libradcli-dev \
     libpaho-mqtt-dev \
+    libcgreen1-dev \
```
y0urself, line 39:
```suggestion
    libcgreen1-dev \
    lcov \
```

y0urself (approved):

PR #638, opened 2021-12-20
Add functions osp_check_feed and osp_get_vts_feed_info

timopollmeier:
**What**:
This adds new functions to get extended information about the NVTs feed via OSP.

**Why**:
So gvmd can get the info from the scanner instead of calling greenbone-nvt-sync directly.
(AP-1845)

**How**:
By running the `<get_feeds/>` GMP command with the changes from greenbone/gvmd/pull/1769.

**Checklist**:

- [ ] Tests

### DIFF
--- a/osp/osp.c
+++ b/osp/osp.c
@@ -249,6 +249,97 @@ osp_connection_close (osp_connection_t *connection)
 }
 
 /**
+ * @brief Gets additional status info about the feed.
+ *
+ * @param[in]   connection            Connection to an OSP server.
+ * @param[out]  lockfile_in_use       Whether the lockfile is in use.
+ * @param[out]  self_test_exit_error  Whether the sync script self check failed.
+ * @param[out]  self_test_error_msg   Self check error message if one occurred.
+ * @param[out]  cmd_error             Error message of the OSP command.
+ *
+ * @return 0 if success, 1 if error.
+ */
+int
+osp_check_feed (osp_connection_t *connection, int *lockfile_in_use,
+                int *self_test_exit_error, char **self_test_error_msg,
+                char **cmd_error)
+{
+  entity_t entity, feed, lockfile_entity, exit_error_entity, error_msg_entity;
+  const char *status, *status_text;
+
+  if (!connection)
+    return 1;
+
+  if (osp_send_command (connection, &entity, "<check_feed/>"))
+    return 1;
+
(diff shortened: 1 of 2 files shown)


osp/osp.c
```diff
@@ -248,6 +248,97 @@ osp_connection_close (osp_connection_t *connection)
...
+  error_msg_entity = entity_child (feed, "self_test_error_msg");
+
+  if (lockfile_in_use)
+    {
+      if (lockfile_entity)
+        *lockfile_in_use = atoi (entity_text (lockfile_entity));
+      else
+        {
+          g_warning ("%s: element LOCKFILE_IN_USE missing.", __func__);
+          *lockfile_in_use = 0;
```
jjnicola, line 307: How do you differentiate the lockfile_in_use not in used from the one with an error? Wouldn't be better to set -1 in case of missing element?

osp/osp.c
```diff
@@ -248,6 +248,97 @@ osp_connection_close (osp_connection_t *connection)
...
+    }
+
+  if (self_test_exit_error)
+    {
+      if (exit_error_entity)
+        *self_test_exit_error = atoi (entity_text (exit_error_entity));
+      else
+        {
+          g_warning ("%s: element SELF_TEST_EXIT_ERROR missing.", __func__);
+          *self_test_exit_error = 0;
```
jjnicola, line 318: Same as above. Exit code 0 is that the script finished successfully. Shouldn't be set with another value for a missing element?

jjnicola (changes requested):

osp/osp.c
timopollmeier: I thought it might be simpler to handle it this way in gvmd but in retrospect it makes more sense to distinguish this error from the lock not being in use in a library function.

osp/osp.c
timopollmeier: Again, I chose this based on the use in gvmd but your suggestion makes more sense for a library function.

jjnicola (approved):

osp/osp.c
jjnicola: solved with 8e59daccdd57c8371ae9d88c50e3d3e3d749c36a

osp/osp.c
jjnicola: solved with 8e59daccdd57c8371ae9d88c50e3d3e3d749c36a

PR #642, opened 2022-01-13
Fix and update the doc CI

y0urself:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests


ArnoStiefvater (approved):

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Upload doxygen error output as artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: doxygen-stderr.txt
+          path: ~/doxygen-stderr.txt
+      - name: Upload XML documentation as artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: xml-doc
+          path: build/doc/generated/xml/
```
bjoernricks, line 38: Would be good to set a retention-days value. For example `retention-days: 7` to store the content only 7 days.

PR #656, opened 2022-02-15
replace strncpy

nichtsfrei:
- Refactor: ignore -Wstringop-truncation due to FOTIFY=2
- Refactor: remove -Werror
- Refactor: enable -Werror replace strncpy with memcpy

### DIFF
--- a/base/prefs.c
+++ b/base/prefs.c
@@ -174,7 +174,8 @@ prefs_config (const char *config)
   if (!global_prefs)
     prefs_init ();
 
-  strncpy (buffer, config, sizeof (buffer));
+  memset (buffer, 0, sizeof (buffer));
+  memcpy (buffer, config, sizeof (buffer));
   if (!init_settings_iterator_from_file (&settings, buffer, "Misc"))
     {
       while (settings_iterator_next (&settings))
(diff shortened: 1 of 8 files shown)


base/prefs.c
```diff
@@ -174,7 +174,8 @@ prefs_config (const char *config)
   if (!global_prefs)
     prefs_init ();
 
-  strncpy (buffer, config, sizeof (buffer));
+  memset (buffer, 0, sizeof (buffer));
+  memcpy (buffer, config, sizeof (buffer));
```
y0urself, line 178:
```suggestion
  memset (buffer, 0, sizeof (buffer));
  memcpy (buffer, config, sizeof (buffer) - 1);
```

we might want to have a null termination at the end?

I have seen in the other places the usage of `bzero` ... we could use it here too?

nichtsfrei: I would rather use the standard memset tbh

osp/osp.c
```diff
@@ -152,7 +152,8 @@ osp_connection_new (const char *host, int port, const char *cacert,
         }
 
       addr.sun_family = AF_UNIX;
-      strncpy (addr.sun_path, host, sizeof (addr.sun_path) - 1);
+      memset (addr.sun_path, 0, sizeof (addr.sun_path) - 1);
```
y0urself, line 155:
```suggestion
      memset (addr.sun_path, 0, sizeof (addr.sun_path));
```

I guess we want to initialize everything?

y0urself (approved):
changes look good to me

PR #674, opened 2022-04-06
Fix: instead of setting strlen of argc just assume _POSIX_PATH_MAX 256 

nichtsfrei:
In some scans it can happen that the program name will go over 256 may
resulting in segmentation faults. To counter that we set the maximal
program name to 256 to avoid malloc or segmentation faults.

### DIFF
--- a/base/proctitle.c
+++ b/base/proctitle.c
@@ -42,7 +42,6 @@ extern const char *__progname;
 #ifndef __FreeBSD__
 extern const char *__progname_full;
 #endif
-static int argv_len;
 static char **old_argv;
 extern char **environ;
 void *current_environ = NULL;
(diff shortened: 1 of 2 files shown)


base/proctitle.c
```diff
@@ -115,15 +109,11 @@ proctitle_set_args (const char *new_title, va_list args)
...
-      i = argv_len - 2;
-      formatted[i] = '\0';
-    }
-  bzero (old_argv[0], argv_len);
-  memcpy (old_argv[0], formatted, argv_len);
-  old_argv[1] = NULL;
+  // _POSIX_PATH_MAX is conservatively defined as 256 so we just assume that
+  // max size in order to not run into segmentation faults or having to malloc
+  memset (old_argv[0], 0, 256);
+  // we ommit one to be 0 terminated
```
Kraemii, line 115:
```suggestion
  // we omit one to be 0 terminated
```

Kraemii (approved):

ArnoStiefvater (approved):

PR #693, opened 2022-07-19
Refactor: adapt workflows from main to stable

nichtsfrei:


nichtsfrei:
This has a chicken-and-egg problem: it cannot run the tests because the stable images aren't updated without this PR and therefore are missing paho library

bjoernricks (approved):

Kraemii (approved):

PR #703, opened 2022-08-23
Change: Split building build and production container images

bjoernricks:
**What**:

Split building build and production container images

**Why**:

The build image really rarely needs a rebuild. Therefore split this job
into an own workflow and only rebuild if the corresponding dockerfile or
workflow has changed.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -1,25 +1,34 @@
-name: Container Image Builds
+name: Build Container Image Builds
 
 on:
   push:
     branches: [ main, stable, oldstable ]
     tags: ["v*"]
+    paths:
+      - .github/workflows/build-container.yml
+      - .docker/build.Dockerfile
   pull_request:
     branches: [ main, stable, oldstable ]
+    paths:
+      - .github/workflows/build-container.yml
+      - .docker/build.Dockerfile
   workflow_dispatch:
+  schedule:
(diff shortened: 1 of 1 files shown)


Dexus:
I would recommend to rebuild the build image at least once all 2 weeks via schedule, because new packages will else not used for the builds.

bjoernricks:
> I would recommend to rebuild the build image at least once all 2 weeks via schedule, because new packages will else not used for the builds.

I've also thought about that. Could you check if the schedule fits now? not sure if i got that cron syntax right

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,65 @@
...
+      - .docker/build.Dockerfile
+  pull_request:
+    branches: [ main, stable, oldstable ]
+    paths:
+      - .github/workflows/build-container.yml
+      - .docker/build.Dockerfile
+  workflow_dispatch:
+  schedule:
+    # rebuild image every sunday
+    - cron: 0 0 * * 0
```
Dexus, line 18:
```suggestion
    - cron: '0 0 * * 0'
```

Dexus (changes requested):

Dexus (approved):

Dexus:
LGTM

PR #708, opened 2022-09-29
Change: Use new trigger-workflow action to update dependent images

bjoernricks:
**What**:

Use new trigger-workflow action to update dependent images

DEVOPS-404

**Why**:

When the gvm-libs container has re-build all dependent images should be re-build too.

**How**:

**Checklist**:

- [ ] Tests


y0urself (approved):
🚀

PR #723, opened 2022-12-06
Set LDAP version to 3 after initialization

fklajn:
LDAP version is by default set to 3, but when StarTLS is disabled then the initialization is done for the second time. And after the initialization version is set to 2. This change is to set the version back to 3.

Signed-off-by: Franciszek Klajn <fklajn@opera.com>

**What**:

LDAP version setting bugfix
jira: T4-398

**Why**:

Because without the change LDAP authentication over LDAPS is not possible on recent Linux distribution.

**How**:

The change has been tested on a test system (Debian Bullseye).

**Checklist**:

- [X] Tests
- [X] PR merge commit message adjusted

### DIFF
--- a/util/ldaputils.c
+++ b/util/ldaputils.c
@@ -333,8 +333,29 @@ ldap_auth_bind (const gchar *host, const gchar *userdn, const gchar *password,
                   g_free (ldapuri);
                   goto fail;
                 }
+             // Set LDAP version to 3 after initialization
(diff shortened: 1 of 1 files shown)


ArnoStiefvater:
Hey
Thanks a lot for your contribution.
It seems that you still have to run `clang-format -i -style=file util/ldaputils.c` to pass the formatting checks.

ArnoStiefvater (approved):

PR #735, opened 2023-02-03
Add: the possibility to set reference value for logs

Kraemii:
**What**:
To be able to detect corresponding log messages it is possible to set a reference value, which is then displayed in logs afterwards.
SC-768

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/base/logging.c
+++ b/base/logging.c
@@ -49,6 +49,12 @@
 #define G_LOG_DOMAIN "libgvm base"
 
 /**
+ * @brief for backward compatibility
+ *
+ */
+#define LOG_REFERENCES_AVAILABLE
+
+/**
  * @struct gvm_logging_t
  * @brief Logging stores the parameters loaded from a log configuration
  * @brief file, to be used internally by the gvm_logging module only.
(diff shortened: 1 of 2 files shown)


base/logging.c
```diff
@@ -383,6 +383,55 @@ gvm_log_unlock (void)
...
+ * simultaniousely it is possible to detect all log corresponding to the same
+ * scan. The log reference is optional and must be set before calling
+ * setup_log_handlers. The data given must be freed by calling
+ * free_log_reference(). If called multiple times the old reference gets freed
+ * and the new one is set instead.
+ *
+ * @param ref
+ */
+void
+set_log_reference (char *ref)
```
nichtsfrei, line 402: To make deployment of GOS less error prone and since we currently don't have semantic version as of yet maybe it makes sense to add a
```
#define LOG_REFERENCES_AVAILABLE
```

as well so that in openvas we can easily check via

```
#ifdef LOG_REFERENCES_AVAILABLE
set_log_reference(...);
#endif 
``` 
this would make it somewhat downwards compatible until gvm-libs went through.

nichtsfrei (changes requested):

jjnicola (dismissed):

jjnicola (approved):

nichtsfrei (approved):

PR #753, opened 2023-04-05
Generate a hash value for a given string.

jhelmold:
## What
Implemented a function that generates a hash value for an arbitrary string.

## Why
This function is needed for GEA-51 to catch duplicated results in reports.

## References
Addresses GEA-51.

## Checklist

- [ ] Tests

### DIFF
--- a/util/authutils.c
+++ b/util/authutils.c
@@ -231,15 +231,15 @@ get_password_hashes (const gchar *password)
 }
 
 /**
- * @brief Generate a hash value for a given string
+ * @brief Calculate the MD5 hash value for a given string
  *
  * @param string The String to be hashed
  *
- * @return A pointer to a gchar containing the hash value, has to be freed by
- *         the caller.
+ * @return A pointer to a gchar containing the hash value as a hexadecimal
+ *         string, has to be freed by the caller.
  */
 gchar *
-get_hash_value_from_string (const gchar *string)
+get_md5_hash_from_string (const gchar *string)
 {
   g_assert (string);
 
--- a/util/authutils.c
+++ b/util/authutils.c
@@ -231,6 +231,30 @@ get_password_hashes (const gchar *password)
 }
 
 /**
+ * @brief Generate a hash value for a given string
+ *
+ * @param string The String to be hashed
+ *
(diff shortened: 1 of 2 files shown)


util/authutils.c
```diff
@@ -230,6 +230,30 @@ get_password_hashes (const gchar *password)
...
+/**
+ * @brief Generate a hash value for a given string
+ *
+ * @param string The String to be hashed
+ *
+ * @return A pointer to a gchar containing the hash value, has to be freed by
+ *         the caller.
+ */
+gchar *
+get_hash_value_from_string (const gchar *string)
```
timopollmeier, line 242:
```suggestion
get_md5_hash_from_string (const gchar *string)
```

util/authutils.h
```diff
@@ -52,6 +52,9 @@ gvm_auth_init (void);
 int
 gvm_authenticate_classic (const gchar *, const gchar *, const gchar *);
 
+gchar *
+get_hash_value_from_string (const gchar *);
```
timopollmeier, line 56:
```suggestion
get_md5_hash_from_string (const gchar *);
```

util/authutils.c
```diff
@@ -230,6 +230,30 @@ get_password_hashes (const gchar *password)
   return hashes_out;
 }
 
+/**
+ * @brief Generate a hash value for a given string
```
timopollmeier, line 234:
```suggestion
 * @brief Calculate the MD5 hash value for a given string
```

util/authutils.c
```diff
@@ -230,6 +230,30 @@ get_password_hashes (const gchar *password)
...
+ *
+ * @param string The String to be hashed
+ *
+ * @return A pointer to a gchar containing the hash value, has to be freed by
+ *         the caller.
```
timopollmeier, line 239:
```suggestion
 * @return A pointer to a gchar containing the hash value as a hexadecimal
 *         string, has to be freed by the caller.
```

timopollmeier (changes requested):
I think it should be more obvious what kind of hash is calculated (MD5) and how it is encoded.

timopollmeier (approved):

PR #771, opened 2023-05-11
Add: gvm_hosts_allowed_only()

jjnicola:
## What
Add: gvm_hosts_allowed_only()
Jira: SC-837

## Why
This function removes denied hosts and kept the allowed hosts from the hosts list.

For backward compatibility, a pre-procesor directive defines a flag which enabled this feature.

## Checklist

- [ ] Tests


base/hosts.h
```diff
@@ -28,6 +28,8 @@
 #ifndef _GVM_HOSTS_H
 #define _GVM_HOSTS_H
 
+#define FEATURE_HOSTS_ALLOWED_ONLY 1
```
ArnoStiefvater, line 31: Briefly document why this is needed and what it does in code.

base/hosts.c
```diff
@@ -1636,6 +1637,118 @@ gvm_hosts_exclude_with_max (gvm_hosts_t *hosts, const char *excluded_str,
   return excluded;
 }
 
+/**
+ * @brief Returns a list of hosts after an host autherization check.
```
4D617274696E, line 1641:
```suggestion
 * @brief Returns a list of hosts after a host authorization check.
```

base/hosts.h
jjnicola: I added above this line. Therefore, it is not shown as `outdated`

Kraemii:
```suggestion
#define FEATURE_HOSTS_ALLOWED_ONLY
```
You also don't need to set a value

nichtsfrei: Although it is true that it isn't required when using `#ifdef` I think it is more clear to set it to 1 since that also allows runtime if like:
```
if (#FEATURE_HOSTS_ALLOWED)
   ...
```

Kraemii (approved):

PR #774, opened 2023-05-23
Fix: Parse XML with XML_PARSE_HUGE option

timopollmeier:
## What
The libxml based parse_element function now parses the input with relaxed limits by using the XML_PARSE_HUGE option. 

## Why
This fixes problems with parsing larger XML files like the CPE dictionary.

## References
GEA-208


timopollmeier:
I haven't found an exact reason either but I think the limits are just there to limit the use of resources, especially memory, in case a large portion of XML is passed to the parser by accident or maliciously from an untrusted source.

jhelmold (approved):

PR #778, opened 2023-06-14
Change: Use debian:bullseye-slim base images

timopollmeier:
## What
Use debian:bullseye-slim base images instead of  debian:stable-slim

## Why
So the release of Debian Bookworm does not break compatibility.

## Checklist

- [ ] Tests


nichtsfrei:
Leaving a comment to block merging as this is not decided yet.

nichtsfrei:
There is now the tag:

- `greenbone/gvm-libs:oldstable-edge`

that provides this functionality.

PR #798, opened 2023-10-17
Change: Don't perform alive tests if "consider alive" method is selected

jjnicola:
## What
When more than one alive test method is selected, boreas check for each one until the host is alive. This will be done also even if the "consider alive" methods is selected. 
Jira: SC-934

## Why
This patch improves boreas to not run other selected alive methods if "consider alive" method is set

## Checklist

- [ ] Tests

### DIFF
--- a/boreas/alivedetection.c
+++ b/boreas/alivedetection.c
@@ -76,6 +76,29 @@ scan (alive_test_t alive_test)
   g_message ("Alive scan %s started: Target has %d hosts", scan_id,
              number_of_targets);
 
+  /* Check first if consider alive test method is set. In case that
+   * there is no scan restrictions, no other test will be performed, because
+   * it doesn't make sense.
+   * Otherwise, if a scan restriction is set (the max_scan_hosts feature),
+   * the other selected methods must be performed first to fill a list with
+   * alive hosts and not only with the first which we don't now if they are dead
+   * or alive. This will ensure that restricted users will get the max possible
+   * amount of alive hosts before starting to fill the list with dead hosts.
+   */
+  if (alive_test & ALIVE_TEST_CONSIDER_ALIVE
+      && scanner.scan_restrictions->max_scan_hosts == 0)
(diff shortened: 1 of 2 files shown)


boreas/boreas_io.c
```diff
@@ -474,6 +474,12 @@ get_alive_test_methods (alive_test_t *alive_test)
     {
       *alive_test = atoi (alive_test_pref_as_str);
     }
+
+  // If consider_alive method is set, unset the other methods
```
ArnoStiefvater, line 478: Not sure if this is the cleanest solution. Feels a bit like hidden logic to place it here.
Other solution would be to handle the ALIVE_TEST_CONSIDER_ALIVE first in the ping functions and just abort after or not start the others.
We can discuss tomorrow.

boreas/alivedetection.c
```diff
@@ -76,6 +76,25 @@ scan (alive_test_t alive_test)
   g_message ("Alive scan %s started: Target has %d hosts", scan_id,
              number_of_targets);
 
+  /* Check first if consider alive test method is set. In case there
+   * that there is no scan restrictions, no other test will be performed.
```
cfi-gb, line 80:
```suggestion
  /* Check first if consider alive test method is set. In case that
   * there is no scan restrictions, no other test will be performed.
```

boreas/alivedetection.c
```diff
@@ -76,6 +76,29 @@ scan (alive_test_t alive_test)
   g_message ("Alive scan %s started: Target has %d hosts", scan_id,
              number_of_targets);
 
+  /* Check first if consider alive test method is set. In case that
+   * there is no scan restrictions, no other test will be performed, because it
```
ArnoStiefvater, line 80:
```suggestion
   * there is no scan restrictions, no other test will be performed, because
```

ArnoStiefvater (approved):

PR #806, opened 2024-01-19
Add: missing dependency in container prod-testing

jjnicola:
## What
Add: missing dependency in container prod-testing

## Why
libhiredis-dev is missing

## Checklist

- [ ] Tests


mattmundell (approved):

bjoernricks:
Sorry but why is a dev dependency missing in a production container?

y0urself:
btw.: I want to also ask the question from @bjoernricks ... why is there a dev dependency missing in the production image?
This should be added to build time not to the run time of the container ...

PR #808, opened 2024-03-01
Add: New XML file iterator for large list-like XML

timopollmeier:
## What
This adds functions and data type for iterating over subelements in an XML file at a given tree depth without having to keep the object model (DOM) of the whole document in memory.

## Why
This is useful for handling larger list-like XML documents like the CPE dictionary.

## References
GEA-488


util/xmlutils.c
```diff
@@ -2317,3 +2317,727 @@ print_element_to_string (element_t element, GString *string)
...
+
+/**
+ * @brief Initializes an XML file iterator to read from a given path.
+ *
+ * @param[in]  iterator         Pointer to the iterator to initialize.
+ * @param[in]  file_path        Path to the file to read from.
+ * @param[in]  output_depth     XML tree depth at which to return elements.
+ *
+ * @return -1 error, 0 success, 1 already initialized,
+ *         2 error opening file (errno is set to reason)
```
a-h-abdelsalam, line 2866:
```suggestion
 *         2 error opening file (errno is set to reason)
 *         3 error creating parser context
```

util/xmlutils.c
```diff
@@ -2317,3 +2317,727 @@ print_element_to_string (element_t element, GString *string)
...
+  if (iterator->file == NULL)
+    return 2;
+
+  iterator->element_queue = g_queue_new ();
+
+  iterator->file_path = g_strdup (file_path);
+
+  xml_file_iterator_init_sax_handler (&(iterator->sax_handler));
+  iterator->parser_ctxt = xmlCreatePushParserCtxt (
+    &(iterator->sax_handler), iterator, NULL, 0, iterator->file_path);
```
a-h-abdelsalam, line 2896: It could be a good idea to check for `xmlCreatePushParserCtxt` returning NULL
```suggestion
  iterator->parser_ctxt = xmlCreatePushParserCtxt (
    &(iterator->sax_handler), iterator, NULL, 0, iterator->file_path);
  if (iterator->parser_ctxt == NULL)
    return 3;
```
If so, the new error code would need to be checked in gvmd.

util/xmlutils.c
```diff
@@ -2317,3 +2317,727 @@ print_element_to_string (element_t element, GString *string)
...
+    {
+      g_queue_clear_full (iterator->element_queue,
+                          (GDestroyNotify) (element_free));
+    }
+
+  if (iterator->parser_ctxt)
+    {
+      xmlFreeParserCtxt (iterator->parser_ctxt);
+      iterator->parser_ctxt = xmlCreatePushParserCtxt (
+        &(iterator->sax_handler), iterator, NULL, 0, iterator->file_path);
```
a-h-abdelsalam, line 2962: Maybe also for this call.

util/xmlutils.c
```diff
@@ -2317,3 +2317,727 @@ print_element_to_string (element_t element, GString *string)
...
+}
+
+/**
+ * @brief Initializes a xmlSAXHandler data structure for SAX version 2 parsing,
+ *        assigning all the XML file iterator callback functions.
+ *
+ * @param[in] hdlr  The xmlSAXHandler to initialize
+ */
+static void
+xml_file_iterator_init_sax_handler (xmlSAXHandler *hdlr)
```
a-h-abdelsalam, line 2810:
```suggestion
xml_file_iterator_init_sax_handler (xmlSAXHandlerPtr hdlr)
```

PR #809, opened 2024-03-15
Fix: Revert "Change: Drop parsing for $Date"

nichtsfrei:
Given testing with the enterprise feed it seems that $Date is still
used.

This reverts commit 32d9e66c351ac0ecab0ea7f6a943db8bd55f4617.


nichtsfrei:
The report was done on an outdated feed.

PR #813, opened 2024-04-16
Change: boreas. Use sequence number 1.

### LINKED ISSUE #812: Boreas use of ICMP Echo with 0 ID can cause issues on some networks
### Expected behavior

ICMP Echos from Boreas should use non zero id and seq. aka "Identifier" and "Sequence Number" from RFC 792 ICMP Echo. And see targets as alive.

Boreas uses identifier=0 seq=0 for ipv4 [ICMP Echo requests during alive checks](https://github.com/greenbone/gvm-libs/blob/main/boreas/ping.c#L183). Some environments appear to not like this. Despite being allowed by the [spec](https://datatracker.ietf.org/doc/html/rfc792), I think this behaviour is possibly at the least confusing since most ping tools increment the  seq by one per message starting from 1, and have identifier > 0

[Ipv6 echos appear to use 234 as the identifier.](https://github.com/greenbone/gvm-libs/blob/main/boreas/ping.c#L147C1-L147C25)

This causes some environments to have all alive test fail. We are moving to "Consider Alive" however we think it benefits others to report this issue.

`ping` (on Ubuntu) appears to pick seq 1..n and identifier=some number above 0 thats from the kernel I believe.

When trying to debug this problem someone may attempt to use `ping` and because it uses a non 0 identifier and seq it will be difficult to debug when compared to the behaviour exhibited by boreas.

### Actual behavior

The ICMP Echo for alive test uses identifier=0 seq=0.

### Steps to reproduce

Start a scan which uses the default alive check of ICMP Echo.
Use tcpdump on the target. As such `tcpdump -v icmp and (src 192.168.1.1 or dst 192.168.1.1) &`

You will see 

```SENT (1.0342s) ICMP [192.168.1.30 > 192.168.1.1 Echo request (type=8/code=0) id=0 seq=0] IP [ttl=64 id=25987 iplen=28 ]```
You will not receive a response if the environment does not appear to allow identifier = 0 Echos.

### GVM versions

I recognize these are older but given the code appears to be unchanged in the last 3-4 years I think the problem still exists.

**gsa:** (gsad --version)
We do not use gsa

**gvm:** (gvmd --version)
Greenbone Vulnerability Manager 21.4.5~git-f27009b24-HEAD

**openvas:** (openvas --version)
OpenVAS 21.4.0~git-32ad87a4-HEAD

**gvm-libs:**
gvm-libs 21.4.4~git-59c8402c-HEAD

### Environment

**Operating system:**
```
uname -a
Linux <HOSTNAME> 5.4.0-148-generic #165-Ubuntu SMP Tue Apr 18 08:53:12 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.6 LTS"
```

**Installation method / source:** (packages, source installation)

We run a mostly unmodified source install. 

### Logfiles

```
libgvm boreas:  DEBUG:2024-04-12 22h38.28 utc:1055519: alive_detection_init: Initialise alive scanner.
libgvm boreas:  DEBUG:2024-04-12 22h38.28 utc:1055519: alive_detection_init: Initialisation of alive scanner finished.
libgvm boreas:MESSAGE:2024-04-12 22h38.28 utc:1055519: Alive scan 07f6713f-3b46-4edb-b282-948aecfb4079 started: Target has 1 hosts
libgvm boreas:  DEBUG:2024-04-12 22h38.30 utc:1055519: scan: all ping packets have been sent, wait a bit for rest of replies.
libgvm boreas:  DEBUG:2024-04-12 22h38.31 utc:1055519: stop_sniffer_thread: Try to stop thread which is sniffing for alive hosts.
libgvm boreas:  DEBUG:2024-04-12 22h38.33 utc:1055519: stop_sniffer_thread: pthread_join() returned PTHREAD_CANCELED.
libgvm boreas:  DEBUG:2024-04-12 22h38.33 utc:1055519: stop_sniffer_thread: Stopped thread which was sniffing for alive hosts.
libgvm boreas:MESSAGE:2024-04-12 22h38.33 utc:1055519: Alive scan 07f6713f-3b46-4edb-b282-948aecfb4079 finished in 5 seconds: 0 alive hosts of 1.
libgvm boreas:  DEBUG:2024-04-12 22h38.34 utc:1055519: get_host_from_queue: Boreas already finished scanning and we reached the end of the Queue of alive hosts.
```

### PR DESCRIPTION
**Pull request by jjnicola**
## What
Fix: boreas. Use sequence number 1
Close #812
Jira: SC-1064

## Checklist

- [ ] Tests

### DIFF
--- a/boreas/ping.c
+++ b/boreas/ping.c
@@ -54,6 +54,16 @@ struct pseudohdr
   struct tcphdr tcpheader;
 };
 
+// Return a random uint16 avoiding the 0.
+static uint16_t
+get_echo_id (void)
+{
+  uint16_t upper_bound = 65534;
+  uint16_t lower_bound = 1;
+
+  return rand () % (upper_bound - lower_bound + 1) + lower_bound;
+}
+
 /**
  * @brief Get the size of the socket send buffer.
  *
@@ -144,8 +154,8 @@ send_icmp_v6 (int soc, struct in6_addr *dst, int type)
   icmp6 = (struct icmp6_hdr *) sendbuf;
   icmp6->icmp6_type = type; /* ND_NEIGHBOR_SOLICIT or ICMP6_ECHO_REQUEST */
   icmp6->icmp6_code = 0;
-  icmp6->icmp6_id = 234;
-  icmp6->icmp6_seq = 0;
+  icmp6->icmp6_id = get_echo_id ();
+  icmp6->icmp6_seq = 0x0100;
 
   memset ((icmp6 + 1), 0xa5, datalen);
   gettimeofday ((struct timeval *) (icmp6 + 1), NULL); // only for testing
@@ -197,6 +207,8 @@ send_icmp_v4 (int soc, struct in_addr *dst)
   icmp = (struct icmphdr *) sendbuf;
   icmp->type = ICMP_ECHO;
   icmp->code = 0;
+  icmp->un.echo.id = get_echo_id ();
+  icmp->un.echo.sequence = 0x0100;
 
   len = 8 + datalen;
   icmp->checksum = 0;
--- a/.docker/prod-testing.Dockerfile
+++ b/.docker/prod-testing.Dockerfile
@@ -21,7 +21,7 @@ RUN apt-get update && \
   libgnutls30 \
   libuuid1 \
   libssh-gcrypt-4 \
-  libhiredis0.14 \
+  libhiredis1.1.0 \
   libhiredis-dev \
   libxml2 \
   libpcap0.8 \

### CODE REVIEW

File: boreas/ping.c [hunk 1]
```diff
@@ -197,6 +197,7 @@ send_icmp_v4 (int soc, struct in_addr *dst)
   icmp = (struct icmphdr *) sendbuf;
   icmp->type = ICMP_ECHO;
   icmp->code = 0;
+  icmp->un.echo.sequence = 0x0100;
```
co60ca, line 200: Should we also populate id?

I'm not a C developer but the value of sendbuf is undefined right?
```
struct icmphdr
{
  u_int8_t type;                /* message type */
  u_int8_t code;                /* type sub-code */
  u_int16_t checksum;
  union
  {
    struct
    {
      u_int16_t        id;
      u_int16_t        sequence;
    } echo;                        /* echo datagram */
    u_int32_t        gateway;        /* gateway address */
    struct
    {
      u_int16_t        __unused;
      u_int16_t        mtu;
    } frag;                        /* path mtu discovery */
  } un;
};
```

boreas/ping.c
```diff
@@ -144,8 +144,8 @@ send_icmp_v6 (int soc, struct in6_addr *dst, int type)
   icmp6 = (struct icmp6_hdr *) sendbuf;
   icmp6->icmp6_type = type; /* ND_NEIGHBOR_SOLICIT or ICMP6_ECHO_REQUEST */
   icmp6->icmp6_code = 0;
-  icmp6->icmp6_id = 234;
-  icmp6->icmp6_seq = 0;
+  icmp6->icmp6_id = rand ();
```
co60ca, line 147: `max(rand (), 1)` ?

That way we can't accidentally get 0 if rand () returns 0? (I'm assuming that rand here can return 0, but again I'm not much of a C developer :P

co60ca (commented):
Hope you don't mind my review. I'm not well versed in C so please take my review with the utmost of respect.

boreas/ping.c
jjnicola: Thanks for the review!

mattmundell (dismissed):

boreas/ping.c
co60ca: I wonder why in our case it was always zero? Are older kernels leaving it as is?
In 
```Linux <HOSTNAME> 5.4.0-148-generic #165-Ubuntu SMP Tue Apr 18 08:53:12 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux``` we were seeing the 

```
SENT (1.0342s) ICMP [192.168.1.30 > 192.168.1.1 Echo request (type=8/code=0) id=0 seq=0] IP [ttl=64 id=25987 iplen=28 ]
```
id=0 for ipv4

cfi-gb:
One note on the usage of "Fix:" for the conventional commit message:

Shouldn't this be "Change:" instead? Asking because when using a "Fix" it will end up as a "Bug Fix" in the Changelog but this is far from being a Bug as the RFC is actually allowing to use a 0/zero:

> If code = 0, a sequence number to aid in matching echos and replies, **may be zero**.
>
> https://datatracker.ietf.org/doc/html/rfc792

and this is more about a compatibility with some specific systems not following the specs.

nichtsfrei (approved):

Kraemii (approved):

PR #814, opened 2024-04-29
Add: Allow getting score of CVSS 4.0 vectors

timopollmeier:
## What
The function get_cvss_score_from_base_metrics can now also calculate the score of CVSS 4.0 vectors.

## Why
To be able to use CVSS 4.0 in VTs.

## References
GEA-529

## Checklist
- [x] Tests

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -70,6 +70,9 @@ if (BUILD_TESTS)
                   EXCLUDE_FROM_ALL
                   array_tests.c)
 
+  add_link_options (-g -lgcov --coverage)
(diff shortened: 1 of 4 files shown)


a-h-abdelsalam (changes requested):
The header of the file `cvss.c `could be updated to include a section for `CVSS v4.0` with links to the specification similar to what exists for other CVSS versions.

a-h-abdelsalam:
There is a code format error that is causing the build CI check to fail. You may need to run: `clang-format -i -style=file {base,boreas,gmp,osp,util}/*.{c,h}
`

a-h-abdelsalam (approved):

PR #822, opened 2024-06-20
Remove libcgreen-dev from build container docker image

bjoernricks:
## What

Remove libcgreen-dev from build container docker image

## Why

libcgreen-dev is not available for debian stable.

## References

https://github.com/greenbone/gvm-libs/actions/runs/9515855202/job/26230948905#step:8:284


timopollmeier:
The build container has been removed, so this is no longer needed.

PR #827, opened 2024-07-11
Add: Added conversion routines for different CPE notation formats.

jhelmold:
## What
Added routines to read a v2.2-conformant URI of a CPE or a formatted string binding of a CPE into some kind of WFN struct. Added routines to convert the WFN struct into a formatted string binding or a CPE v2.2-conformant URI. Added routines to convert a formatted string binding into a v2.2-conformant URI and vice versa.

Also added a function, that takes two CPEs represented in CPE structs and checks if one of the CPEs matches the other one.

## Why
These routines are needed for direct matching.

## References
GEA-622
GEA-624

## Checklist

Added some unit tests.
- [x] Tests

### DIFF
--- a/util/cpeutils.c
+++ b/util/cpeutils.c
@@ -16,10 +16,10 @@
 #include "cpeutils.h"
 
 #include <assert.h>
+#include <ctype.h>
 #include <errno.h>
 #include <glib.h>
 #include <string.h>
-#include <ctype.h>
 
 #undef G_LOG_DOMAIN
 /**
@@ -41,7 +41,7 @@ uri_cpe_to_fs_cpe (const char *uri_cpe)
 
   cpe_struct_init (&cpe);
   uri_cpe_to_cpe_struct (uri_cpe, &cpe);
-  return(cpe_struct_to_fs_cpe (&cpe));
+  return (cpe_struct_to_fs_cpe (&cpe));
 }
 
 /**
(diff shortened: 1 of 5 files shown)


a-h-abdelsalam (changes requested):

util/cpeutils.c
```diff
@@ -0,0 +1,1276 @@
...
+  memset (*dest, 0, size + 1);
+  strncpy (*dest, src, size);
+}
+
+/**
+ * @brief Returns if source is a match for target. That means
+ *        that source is a superset of target.
+ *
+ * @param[in]  source  The cpe_struct that represents a set of CPEs.
+ *             target  The cpe_struct that represents a single CPE or
```
a-h-abdelsalam, line 1026:
```suggestion
 * @param[in]  source  The cpe_struct that represents a set of CPEs.
 * @param[in]  target  The cpe_struct that represents a single CPE or
```

util/cpeutils.c
```diff
@@ -0,0 +1,1276 @@
...
+    return (FALSE);
+  relation = compare_component (source.language, target.language);
+  if (relation != SUPERSET && relation != EQUAL)
+    return (FALSE);
+
+  return (TRUE);
+}
+
+static enum set_relation
+compare_component (const char *source, const char *target)
```
a-h-abdelsalam, line 1076: You may want to add descriptions to this function and following ones.

PR #828, opened 2024-07-25
Add: Added functions used for the matching of CPEs to rules.

jhelmold:
## What
Added a function to compare two version numbers / version strings together with its helper functions.

## Why
This functions are needed for the matching of CPEs to rules.

## References
GEA-624

## Checklist

Implemented some unit tests.
- [x] Tests


util/versionutils.c
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+  if (begin == (int) strlen (version))
+    return (NULL);
+
+  for (end = begin + 1; end < (int) strlen (version) && version[end] != '.';
+       end++)
+    ;
+
+  return (str_cpy ((char *) (version + begin), end - begin));
```
mattmundell, line 293: How about using `g_strdup`?

util/versionutils.c
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+      prep_version[index_pv] = ver[index_v];
+      index_v++;
+      index_pv++;
+    }
+
+  prep_version[index_pv] = '\0';
+  g_free (ver);
+  return (g_strdup (prep_version));
```
mattmundell, line 228: You're using g_strdup so I think you should return a gchar*. I know that in the past we aimed to keep gchar* out of the headers, but many of the newer headers do use gchar*.

Also I recommend dropping the parenthesis around all your return values, to match all the other code.

util/versionutils.c
```diff
@@ -0,0 +1,333 @@
...
+  index_v = index_pv = 0;
+
+  is_digit = g_ascii_isdigit (ver[0]);
+
+  while (index_v < (int) strlen (ver))
+    {
+      if (ver[index_v] == '_' || ver[index_v] == '-' || ver[index_v] == '+'
+          || ver[index_v] == ':' || ver[index_v] == '.')
+        {
+          if (index_pv > 0 && prep_version[index_pv - 1] != '.')
```
mattmundell, line 181: I've only looked briefly but this looks like it could access -1 in prep_version when index_pv is 0.

Also prep_version is of limited size, so I'm guessing in general there should be size checks before it is accessed.

util/versionutils.c
```diff
@@ -0,0 +1,333 @@
...
+  char *ver1, *ver2;
+  char *part1, *part2;
+  int index1 = 0, index2 = 0;
+  int release_state1 = 0, release_state2 = 0;
+  int rs1, rs2;
+
+  ver1 = prepare_version_string (version1);
+  ver2 = prepare_version_string (version2);
+
+  if (strcmp (ver1, ver2) == 0)
```
mattmundell, line 53: Bit risky to pass these straight to strcmp as they may be NULL?

mattmundell (dismissed):

a-h-abdelsalam (approved):

util/versionutils.c
mattmundell: There is [g_strndup](https://docs.gtk.org/glib/func.strndup.html). Thanks for the changes.

PR #852, opened 2024-11-25
Change: move gvm_json_string_escape to a new file

mattmundell:
## What

Move the utility function `gvm_json_string_escape` to a new file.

## Why

This is the start of collecting generic JSON utilities in a dedicated file.

Everything else in `jsonpull.c` is specific to the JSON pull facility. This also helps reduce the size of `jsonpull.c`.

The idea is to put more generic functions into `json.c` in the future. For example `openvasd.c` and `vtparser.c` are using a lot of repeated cJSON code snippets to access doubles. These snippets could be replaced by a `gvm_json_obj_double` function in `json.c`.

This also will help to centralize the use of cJSON. 

## Checklist

I've tested running an `ospd` scan with the new scanner. Seems fine.


bjoernricks (approved):
looks good to me

PR #853, opened 2024-11-28
Fix: openvasd: get result detail source from right place

mattmundell:
## What

In `openvasd_parsed_results`, get `detail_source_*` from `detail > source > *`, instead of from  `detail > value > *`.

## Why

This seems to be the intention, according to the API and rust code.

Note that this also catches a problem where `cJSON_GetObjectItem (detail_obj, "type")` could have been called when `detail_obj` was NULL.


mattmundell:
Hey @jjnicola, saw this while trying some cleanup in #852. This is what you had in mind, right? It compiles, but I still have to set up the JSON things. Maybe you can test?

bjoernricks:
As @mattmundell wrote it would be nice to get tests for this code. Without tests this code is prone to errors like this one.

mattmundell:
After some digging through the code, I now have the Rust scanner set up and I've confirmed that this fits with a JSON based scan. So it's good to go.

I've also added a test case for host details.

PR #867, opened 2025-01-12
Shorten overlength strings instead of downgrading warning

mattmundell:
## What

Literal strings in the code that are longer than 4095 lead to a warnings.

In the networking test, use allocation to shorten the strings in the source code, instead of pushing a diagnostic.

## Why

Pushing a diagnostic prevents the warnings from becoming an error, but the warnings still appear in the compile output.

I know it's perverse to adjust so much for a warning, but I keep getting thrown by seeing the warning in the output even though I have -Werror.


mattmundell:
If you open the patch in a good tool (eg Emacs) you can see that the port numbers all stay the same, and only space and `,` are added around the numbers. Github basically just shows per-line added and removed.

shot

bjoernricks (approved):

PR #871, opened 2025-01-17
Fix: remove stray json-glib config

mattmundell:
## What

Remove `GLIB_JSON_*` references from CMakeLists.txt.

## Why

We're using cJSON now.

## References

Missed in /pull/863.


bjoernricks (approved):

bjoernricks:
Thanks a lot! completely missed these.

PR #874, opened 2025-01-23
Scan start end time

jjnicola:
## What
Fix: start time and end time when parsing scan status
Also, split a function for testing purposes. Add a test.

## Why
 Was always set to 0 because a wrong comparison

## Checklist

- [x] Tests


mattmundell (approved):
Nice.
Would make review a little easier if you split the formatting into a separate commit.

PR #876, opened 2025-01-23
Add: gvm_json_obj_double

mattmundell:
## What

Add `gvm_json_obj_double` to `json.c`, and use it in `openvasd.c` and `vtparser.c`.

## Why

This is neater than using the three line cJSON snippet.

A nice side effect is that this removes many assignments that were happening inside `if` conditions.

## References

Related: /pull/852

## Checklist

- [x] Tests


bjoernricks (approved):
LGTM

PR #877, opened 2025-01-24
Add: gvm_json_obj_str

mattmundell:
## What

Add `gvm_json_obj_str` to `json.c`, and use it in `openvasd.c` and `vtparser.c`.

## Why

This is neater than using the three line cJSON snippet.

As with `gvm_json_obj_double`, this removes many assignments that were happening inside `if` conditions.

## References

Similar to /pull/876.

## Checklist

- [x] Tests


bjoernricks (approved):
Indeed, looks nicer then before

PR #882, opened 2025-02-05
Add: gvm_json_obj_check_str

mattmundell:
## What

Add function `gvm_json_obj_check_str` and use it in `vtparser.c` and `openvasd.c`.

This is a variation on `gvm_json_obj_str` that allows you to retrieve the string field and at the same time get a return value that indicates if the field existed.

## Why

Neater than using the repeated cJSON snippets.

As with the similar PRs, this has the nice side effect of removing assignments from `if` conditions.

Note that the function returns 0 if the field exists, and an error value otherwise. Currently the error is always 1, but this style allows us to add more detail to the error value in future, eg if someone wants to know if the field existed but was the wrong type it could return 2.

## References

Similar to /pull/881 for ints.
Follows `gvm_json_obj_str` from /pull/877.

## Checklist

- [x] Tests

### DIFF
--- a/openvasd/openvasd.c
+++ b/openvasd/openvasd.c
@@ -1383,7 +1383,6 @@ static int
 parse_status (const gchar *body, openvasd_scan_status_t status_info)
 {
   cJSON *parser = NULL;
-  cJSON *status = NULL;
   gchar *status_val = NULL;
   openvasd_status_t status_code = OPENVASD_SCAN_STATUS_ERROR;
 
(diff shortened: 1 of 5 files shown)


bjoernricks (approved):

mattmundell:
> It seems we don't have tests for vtparser.c yet.

I'll take a look next week.

bjoernricks:
> > It seems we don't have tests for vtparser.c yet.
> 
> I'll take a look next week.

That would be really great! thanks is lot in advance!

PR #883, opened 2025-02-06
Fix: leak in JSON parse_results

mattmundell:
## What

Remove `g_strdup` from all details strings in `parse_results`.

## Why

These strings are already being duplicated inside `openvasd_result_new`, which is called later in that block.


bjoernricks (approved):

bjoernricks:
@mattmundell there is a conflict in openvasd.c :-/

mattmundell:
> @mattmundell there is a conflict in openvasd.c :-/

Because #881 removed the `port` var at the same place this PR is moving the `detail_` vars.

bjoernricks (approved):

PR #884, opened 2025-02-07
 Use standard release workflow for gvm-libs

bjoernricks:
## What

 Use standard release workflow for gvm-libs

## Why

Finally we should get changelogs again.

## References

https://jira.greenbone.net/browse/GEA-857


y0urself (approved):
Much better. ❤️

PR #886, opened 2025-02-11
Improve cmake for testing

bjoernricks:
## What

* Add all tests to `all` make target to allow running the tests in vscode correctly
* Introduce a CMake macro for adding unit tests
* Use add unit test macro for specifying all tests

## Why

Easier adding of tests and allow to run tests in vscode without issues

## References

https://jira.greenbone.net/browse/GEA-905

## Checklist

- [x] Tests


CMakeLists.txt
```diff
@@ -32,6 +32,10 @@ else (NOT BUILD_STATIC)
   set (BUILD_SHARED OFF)
 endif (NOT BUILD_STATIC)
 
+set (CMAKE_MODULE_PATH
+  ${PROJECT_SOURCE_DIR}/cmake
+)
```
mattmundell, line 37: Nit: everywhere else puts the ) on the same line, and aligns multiple lines with the first arg.

mattmundell (dismissed):
Nice.

CMakeLists.txt
bjoernricks: Would be nice to have an auto formatter for cmake...

bjoernricks: https://github.com/BlankSpruce/gersemi looks promising. Maybe we should add it to the CI

mattmundell: Maybe [Gersemi](https://github.com/BlankSpruce/gersemi).

Also there's [a cmake lsp](https://github.com/neocmakelsp/neocmakelsp) that says it has a cli for code formatting.

mattmundell (approved):

CMakeLists.txt
mattmundell: > https://github.com/BlankSpruce/gersemi looks promising. Maybe we should add it to the CI

Would be nice to add something.

PR #894, opened 2025-02-25
Change: move assignments out of if conditions (util)

mattmundell:
## What

Move assignment statements out of `if` conditions where possible. So
```
if ((a = b))...
```
becomes
```
a = b
if (a)...
```

## Why

Putting the `=` in the condition is error prone because it's easy to read the `=` as `==`.

Having the `=` in the condition is also harder to read because an extra pair of parens is required around the `=` statement.

## References

Follow on to /pull/890.

### DIFF
--- a/util/kb.c
+++ b/util/kb.c
@@ -251,7 +251,8 @@ parse_port_of_addr (const char *addr, int tcp_indicator_len)
 {
   char *tmp;
   int is_ip_v6;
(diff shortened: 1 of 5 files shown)


mattmundell:
This touches a few files that are missing tests, I'll add those in separate PRs.

bjoernricks (approved):

PR #907, opened 2025-03-28
Fix: some issues producing a segmentation fault.

jjnicola:
## What
- Initialize response in openvasd_get_scan_status() 
- Do not free the customheader in the openvasd_send_request(), since it must be free by the caller function.

## Why
Crashes

## Checklist

- [] Tests

### DIFF
--- a/openvasd/openvasd.c
+++ b/openvasd/openvasd.c
@@ -299,7 +299,6 @@ openvasd_send_request (openvasd_connector_t conn,
     {
       g_warning ("%s: Error performing CURL request", __func__);
       response->body = g_strdup ("{\"error\": \"Error sending request\"}");
-      gvm_http_headers_free (custom_headers);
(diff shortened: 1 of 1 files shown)


openvasd/openvasd.c
```diff
@@ -914,6 +913,7 @@ openvasd_get_scan_status (openvasd_connector_t conn)
     }
   else
     {
+      response = g_malloc0 (sizeof (struct openvasd_response));
```
mattmundell, line 916: It looks like there are are cases that need this, eg `openvasd_stop_scan`.

And for example `openvasd_get_scan_results` is leaking the result because it allocates it at the top, then `openvasd_send_request` assigns over that.

Happy to do a PR for these.

mattmundell (approved):

bjoernricks (approved):

PR #908, opened 2025-04-01
Change: move customheader into openvasd_send_request

mattmundell:
## What

Create customheader inside `openvasd_send_request`.

## Why

All the `openvasd_send_request` callers create the exact same customheader and then free it directly afterwards, so it's simpler to do it inside `openvasd_send_request`.

### DIFF
--- a/openvasd/openvasd.c
+++ b/openvasd/openvasd.c
@@ -263,7 +263,6 @@ static openvasd_resp_t
 openvasd_send_request (openvasd_connector_t conn,
                        gvm_http_method_t method, const gchar *path,
                        const gchar *data,
-                       gvm_http_headers_t *custom_headers,
                        const gchar *header_name)
 {
   openvasd_resp_t response = g_malloc0 (sizeof (struct openvasd_response));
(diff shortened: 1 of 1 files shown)


openvasd/openvasd.c
```diff
@@ -285,6 +284,8 @@ openvasd_send_request (openvasd_connector_t conn,
       conn->stream_resp = g_malloc0 (sizeof (struct gvm_http_response_stream));
     }
 
+  gvm_http_headers_t *custom_headers = init_customheader (conn->apikey, TRUE);
```
timopollmeier, line 287: I think moving the custom headers here is a good idea but I think the second parameter of `init_customheader` for setting the content type should depend on whether the data is empty or not.

timopollmeier (changes requested):

openvasd/openvasd.c
mattmundell: Oh shoot I missed that. Fixed in [c2d6221](https://github.com/greenbone/gvm-libs/pull/908/commits/c2d6221ee350d304b1b5510ea7825075781e1da8).

PR #912, opened 2025-04-09
Build tests without OPENVASD is enabled

bjoernricks:
## What

Build tests without OPENVASD is enabled

## Why

Some tests are only buildable if OPENVASD is enabled. Therefore build these tests conditionally too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,34 +233,40 @@ if (BUILD_TESTS AND NOT SKIP_SRC)
 
   find_package (cgreen REQUIRED)
 
-  add_custom_target (tests DEPENDS
-                     array-test
-                     boreas-alivedetection-test
-                     boreas-cli-test
-                     boreas-error-test
-                     boreas-io-test
-                     boreas-ping-test
-                     boreas-sniffer-test
-                     compressutils-test
-                     cpeutils-test
-                     cvss-test
-                     hosts-test
-                     httputils-test
-                     json-test
-                     jsonpull-test
-                     logging-test
-                     logging-domain-test
-                     networking-test
-                     nvti-test
-                     openvasd-test
-                     osp-test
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -233,34 +233,42 @@ if (BUILD_TESTS AND NOT SKIP_SRC)
...
-                     osp-test
-                     passwordbasedauthentication-test
-                     test-hosts
-                     util-test
-                     version-test
-                     versionutils-test
-                     vtparser-test
-                     xmlutils-test)
+  set (TESTS
+      array-test
```
mattmundell, line 237: Looks like the indent is one char too little here.

CMakeLists.txt
```diff
@@ -233,34 +233,42 @@ if (BUILD_TESTS AND NOT SKIP_SRC)
...
+      networking-test
+      nvti-test
+      osp-test
+      passwordbasedauthentication-test
+      test-hosts
+      util-test
+      version-test
+      versionutils-test
+      xmlutils-test
+    )
```
mattmundell, line 261: Nitpick: this breaks the bracket convention. Everywhere else is closing on the same line.

CMakeLists.txt
```diff
@@ -233,34 +233,42 @@ if (BUILD_TESTS AND NOT SKIP_SRC)
...
+
+  if (OPENVASD)
+    list (APPEND TESTS
```
mattmundell, line 264: Same here.

ozgen (approved):

CMakeLists.txt
bjoernricks: It seems the formatting is not consistent in all cases in this file. Nevertheless I adapted the formatting as you suggested because it was completely different. I am going to look into an autoformatter tool in near future for cmake files.

PR #917, opened 2025-04-29
Add: Add patch method support to httputils for agent controller

ozgen:
## What

Add patch method support to httputils for agent controller

## Why

The PATCH method is needed to update agent states via agent controller API.

## References

GEA-1046


bjoernricks (commented):
Where do we need patch for agent-control? I don't see it in https://greenbone.github.io/scanner-api/ ?

ozgen:
> Where do we need patch for agent-control? I don't see it in https://greenbone.github.io/scanner-api/ ?

the PATCH method is needed specifically for the agent-control service, not for openvasd.
ref : https://github.com/greenbone/agent-control/blob/main/pkg/web/api.go#L471

PR #920, opened 2025-05-05
Fix: port field in result

jjnicola:
## What
Fix: port field in result
Jira: SC-1294

Also, fix format with clang-format

## Why
While openvasd only handle ports as integers, gvmd still waits for general/Host_Details, general/tcp or general/udp.
 This patch handles all these possibilities and set the result port field correctly.

## Checklist

- [ ] Tests

### DIFF
--- a/openvasd/openvasd.c
+++ b/openvasd/openvasd.c
@@ -156,9 +156,10 @@ openvasd_connector_builder (openvasd_connector_t conn, openvasd_conn_opt_t opt,
       conn->apikey = g_strdup ((char *) val);
       break;
     case OPENVASD_PROTOCOL:
-      if (g_strcmp0((char *) val, "http") != 0 && g_strcmp0((char *) val, "https") != 0)
+      if (g_strcmp0 ((char *) val, "http") != 0
+          && g_strcmp0 ((char *) val, "https") != 0)
         return OPENVASD_INVALID_VALUE;
-      conn->protocol = g_strdup((char *) val);
+      conn->protocol = g_strdup ((char *) val);
       break;
     case OPENVASD_HOST:
       conn->host = g_strdup ((char *) val);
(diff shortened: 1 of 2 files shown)


openvasd/openvasd.c
```diff
@@ -704,15 +702,22 @@ openvasd_result_new (unsigned long id, gchar *type, gchar *ip_address,
...
   result->detail_value = g_strdup (detail_value);
   result->detail_source_name = g_strdup (detail_source_name);
   result->detail_source_type = g_strdup (detail_source_type);
   result->detail_source_description = g_strdup (detail_source_description);
 
+  if (!g_strcmp0 (type, "host_detail"))
+    result->port = g_strdup ("general/Host_Details");
+  else if (port == NULL || (!g_strcmp0 (port, "0") && protocol))
+    result->port = g_strdup_printf ("general/%s", protocol);
+  else if (port && protocol)
```
mattmundell, line 716: The `port` check is redundant, right? It's compared to NULL above.

Hard to review this kind of patch where the formatting changes are mixed in.

jjnicola: Yes, sorry for the formatting. I have run clang-format thinking on my changes, but a lot of code was changed. Next time I will check/fix the format after committing.

openvasd/openvasd.c
```diff
@@ -842,14 +846,14 @@ parse_results (const gchar *body, GSList **results)
               gvm_json_obj_str (detail_obj, "description");
           }
       }
+    port = g_strdup_printf ("%d", gvm_json_obj_int (result_obj, "port")),
```
mattmundell, line 849: Should be freed after.

mattmundell (changes requested):
Looks fine. Just that missing free.

mattmundell (approved):

PR #931, opened 2025-06-04
Change: Revert openvasd scan progress calculation

ozgen:
## What

This PR reverts the recent changes to the scan progress calculation logic in openvasd.

## Why

The previous implementation was inaccurate and always reported the scan progress as 0%.

## References

GEA-937


timopollmeier (approved):

jjnicola:
Changes regarding progress calculation where done with #880 which was accidentally merged, and which depends on a still open PR on the scanner side. Once the scanner PR is finally merged, this changes must be applied again.

PR #946, opened 2025-08-12
 Chore: set fix debian codename in /etc/apt/*.sources

nichtsfrei:
To prevent accidental distribution upgrades, e.g. when using gvm-libs as
a base, we should set the codename instead of rolling release.

This minimizes the effort of keeping track of debian releases manually,
checking always for whatever is stable while having a safe guard for
underlying images or runtime environments.

Depends on: https://github.com/greenbone/gvm-libs/pull/945


nichtsfrei:
But your project, your choice. Please inform us upfront so that we can adjust our pipeline for testing against that as we promise debian stable compatibility within openvas-scanner.

PR #951, opened 2025-08-22
Fix: change comment of the agent_controller_convert_scan_agent_config_string

ozgen:
## What

Replaced `cJSON_free` with `g_free` when releasing the payload to ensure the correct memory deallocation function is used. 

## Why

Using `cJSON_free` for memory allocated with GLib functions caused a mismatch in allocation/free.

## References

GEA-1126

### DIFF
--- a/agent_controller/agent_controller.c
+++ b/agent_controller/agent_controller.c
@@ -1004,7 +1004,7 @@ agent_controller_delete_agents (agent_controller_connector_t conn,
  *
  * @return Newly allocated, unformatted JSON string on success; NULL on failure
  *         The caller owns the returned string and must free it with
- *         g_free().
+ *         cJSON_free().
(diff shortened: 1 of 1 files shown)


mattmundell (changes requested):
`agent_controller_convert_scan_agent_config_string` is returning the result of `cJSON_PrintUnformatted`, so it is correct to use `cJSON_free`.

mattmundell:
If you look at https://github.com/DaveGamble/cJSON/blob/master/cJSON.c#L1307 is calls print which is allocating with `hooks->allocate`. So it should use `cJSON_free` which is using `hooks->deallocate`.

mattmundell:
Sorry for causing churn!

mattmundell (approved):

PR #976, opened 2025-09-16
Fix: a couple cases in cmp_versions

mattmundell:
## What

Fix two errors in `cmp_versions`.

## Why

Certain version checks were failing.

## Checklist

- [x] Tests


mattmundell:
@jhelmold It seems like this is what you had in mind for `cmp_versions`. I've added some tests to confirm.

bjoernricks (approved):

PR #981, opened 2025-09-26
Add: Support optional prefix for scan endpoints.

a-h-abdelsalam:
## What
Support optional prefix for scan endpoints.

## Why
Container scanning requires a prefix.

## References
GEA-1283

### DIFF
--- a/http_scanner/http_scanner.c
+++ b/http_scanner/http_scanner.c
@@ -466,6 +466,26 @@ http_scanner_get_version (http_scanner_connector_t conn)
 }
 
 /**
+ * @brief Builds the path prefix for scan endpoints.
+ *
+ * @param conn Connector struct with the data necessary for the connection.
+ *
+ * @return The constructed path prefix. Has to be freed by the caller.
+ */
+static GString *
+build_path_prefix (http_scanner_connector_t conn)
+{
+  GString *path = g_string_new ("");
+  if (conn != NULL && conn->scan_prefix != NULL && conn->scan_prefix[0] != '\0')
+    {
+      g_string_append (path, "/");
+      g_string_append (path, conn->scan_prefix);
+    }
+  g_string_append (path, "/scans");
+  return path;
+}
+
+/**
  * @Brief Create a scan.
  *
  * @param conn Connector struct with the data necessary for the connection.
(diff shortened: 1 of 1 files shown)


http_scanner/http_scanner.c
```diff
@@ -989,10 +1016,15 @@ http_scanner_resp_t
...
 
-  // Stop the scan
-  path = g_string_new ("/scans");
+  if (conn != NULL && conn->scan_prefix != NULL && conn->scan_prefix[0] != '\0')
+    {
+      g_string_append (path, "/");
+      g_string_append (path, conn->scan_prefix);
+    }
+
+  g_string_append (path, "/scans");
```
mattmundell, line 1027: This is repeated 5 or so times now. How about making a function?

It could take an arg to handle the "/scans/preferences" unusual case below.

a-h-abdelsalam: yes I thought about a function that builds the path but there is more than one special case. In addition to preferences, start and stop scan both need the scan id and get results needs a special handling for the range

mattmundell: I see. They all seem the same up to "/scans", which would save some repetition.

http_scanner/http_scanner.c
```diff
@@ -460,28 +465,43 @@ http_scanner_get_version (http_scanner_connector_t conn)
...
   http_scanner_resp_t response = NULL;
   cJSON *parser = NULL;
-  GString *path;
+  GString *path = g_string_new ("");
 
-  if (prefix && prefix[0] != '\0')
-    path = g_string_new (g_strdup_printf ("/%s/scans", prefix));
-  else
-    path = g_string_new ("/scans");
+  if (conn)
```
mattmundell, line 503: I'm a bit surprised by this, because now path will be empty for `http_scanner_send_request` when `conn` is NULL, whereas before path was still "/scans".

Is it intentional that you want to change the behaviour? I expected you'd just be able to go something like
```
GString *path = build_path (conn);
```
and then path would be ready for `http_scanner_send_request`.

a-h-abdelsalam: I don't think the behavior has really changed. `http_scanner_send_request` checks first thing if the `conn` is NULL and if so returns an error response. Therefore, the path does not make any difference if there is no connection.

However, I do see your point if you think it looks better when the path is set in all cases.

bjoernricks (approved):

PR #1046, opened 2026-02-04
Fix: Add missing slash in CyberArk request URLs

timopollmeier:
## What
In cyberark_send_request a slash is added to the URL, separating the hostname and optional port from the path.

## Why
This makes the connection work without having to add a slash a the start of the path manually.

## References
GEA-1562


mattmundell (approved):
I'm not sure what the callers are, but you could also support both (add the / if it's missing).

PR #1069, opened 2026-05-05
Change: Move common functions for handling target credentials to utils.

a-h-abdelsalam:
## What
 Move common functions for handling target credentials to utils.

## Why
To avoid duplicated functions. Target credentials are handled in the same way for OSP and OpenVASD targets.

## References
GEA-1736

### DIFF
--- a/util/credentialutils_tests.c
+++ b/util/credentialutils_tests.c
@@ -1,0 +1,139 @@
+/* SPDX-FileCopyrightText: 2026 Greenbone AG
+ *
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+#include "credentialutils.c"
+
+#include <cgreen/cgreen.h>
+#include <cgreen/mocks.h>
+#include <fcntl.h>
+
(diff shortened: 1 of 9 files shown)


util/credentialutils_tests.c
```diff
@@ -0,0 +1,139 @@
...
+  assert_that (username, is_equal_to_string ("admin"));
+
+  scan_credential_set_auth_data (credential, "username", NULL);
+  username = scan_credential_get_auth_data (credential, "username");
+  assert_that (username, is_null);
+
+  scan_credential_free (credential);
+}
+
+Ensure (credentialutils, credentialutils_set_auth_data_invalid_name)
```
rdttmr: I personally prefer test-case names like `set_auth_data_rejects_invalid_name`, but this is just personal preference, so not a deal-breaker for merging.

Maybe we should align within the team sometime, how we want to name test cases.

robindittmar (dismissed):

robindittmar (approved):

PR #1072, opened 2026-05-19
Change/agent update list payload

ozgen:
## What

-  Changed agent update payload handling to use a list of agent-specific updates.
-  Updated PATCH payload creation to build updates from `agent_controller_agent_update_list_t`.
-  Added `agent_id` and `base` config handling to each update entry.
-  Updated config merging so missing update values are filled from the base config.
-  Updated and extended unit tests for the new update-list payload flow.
-  Verified with the agent controller unit tests.

## Why

-  The previous update flow used one shared update template for multiple agents.
-  The new structure allows each agent to have its own update values.
-  This makes the payload clearer and closer to the Agent Controller API format.
-  It also makes future changes easier, especially when different agents need different update data.

## References

GEA-1791

## Checklist

- [x] Tests

### DIFF
--- a/agent_controller/agent_controller.c
+++ b/agent_controller/agent_controller.c
@@ -474,10 +474,10 @@ dup_str_ptr_array (const GPtrArray *src)
  */
 static agent_controller_agent_config_t
 agent_controller_build_agent_config_with_defaults (
-  const agent_controller_agent_t agent,
+  const agent_controller_agent_config_t base_cfg,
   const agent_controller_agent_config_t update_cfg)
 {
-  if (!agent || !agent->config)
+  if (!base_cfg)
     return NULL;
 
   agent_controller_agent_config_t merged = agent_controller_agent_config_new ();
@@ -488,7 +488,7 @@ agent_controller_build_agent_config_with_defaults (
   if (update_cfg)
     *merged = *update_cfg;
   else
-    *merged = *agent->config;
+    *merged = *base_cfg;
 
   merged->agent_script_executor.scheduler_cron_time = NULL;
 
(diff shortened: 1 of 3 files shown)


agent_controller/agent_controller.c
```diff
@@ -466,18 +466,18 @@ dup_str_ptr_array (const GPtrArray *src)
  * The returned configuration is heap-allocated and must be freed
  * using agent_controller_agent_config_free().
  *
- * @param[in] agent       Agent providing the base configuration.
- * @param[in] update_cfg  Optional configuration overrides. May be NULL.
+ * @param[in] base_cfg       Agent providing the base configuration.
```
timopollmeier:
```suggestion
 * @param[in] base_cfg       The configuration to create a modified copy of.
```

agent_controller/agent_controller.c
```diff
@@ -466,18 +466,18 @@ dup_str_ptr_array (const GPtrArray *src)
...
+ * @param[in] update_cfg     Optional configuration overrides. May be NULL.
  *
  * @return Newly allocated merged configuration on success,
  *         or NULL on error.
  */
 static agent_controller_agent_config_t
 agent_controller_build_agent_config_with_defaults (
```
timopollmeier: I think this should be renamed to something like `agent_controller_build_updated_agent_config`.
Otherwise it sounds like there are some built-in defaults used.

agent_controller/agent_controller.c
```diff
@@ -466,18 +466,18 @@ dup_str_ptr_array (const GPtrArray *src)
...
-  const agent_controller_agent_t agent,
+  const agent_controller_agent_config_t base_cfg,
   const agent_controller_agent_config_t update_cfg)
```
timopollmeier: I think using `config` instead of `cfg` in a parameter name looks more consistent, and the name of the second parameter should also highlight that it contains only the differences, not a full second config.
```suggestion
  const agent_controller_agent_config_t base_config,
  const agent_controller_agent_config_t config_changes)
```

agent_controller/agent_controller.h
```diff
@@ -189,11 +189,24 @@ struct agent_controller_agent_update
 {
   int authorized;                         ///< Authorization status for update
   int update_to_latest;                   ///< Automatically update the agent
-                                          ///  to the latest available version.
-  agent_controller_agent_config_t config; ///< The Agent scan configuration
+                                          ///  to the latest available version
+  agent_controller_agent_config_t config; ///< The new agent scan config
+  agent_controller_agent_config_t base;   ///< The existing agent scan config
```
timopollmeier: These should use the same names as the parameters of the function that builds the updated config. For me it makes things easier to understand and it clarifies that what's currently `config` is actually just the changes (which should also be reflected in the comment).

agent_controller/agent_controller.c
```diff
@@ -466,18 +466,18 @@ dup_str_ptr_array (const GPtrArray *src)
  * The returned configuration is heap-allocated and must be freed
  * using agent_controller_agent_config_free().
  *
- * @param[in] agent       Agent providing the base configuration.
- * @param[in] update_cfg  Optional configuration overrides. May be NULL.
+ * @param[in] base_cfg       Agent providing the base configuration.
+ * @param[in] update_cfg     Optional configuration overrides. May be NULL.
```
timopollmeier: I'd avoid the term `override` and use the same as in the parameter name.

timopollmeier (approved):

PR #1076, opened 2026-06-01
Add/agent controller installer instructions

ozgen:
## What

- Added support for getting installer instructions from the Agent Controller.
- Added language support for the installer instruction request.
- Added custom headers for this request.
- Fixed URL setup error handling.
- Added unit tests.

## Why

- We need to request installer instructions in the selected language.
- The request needs an `Accept: application/json` header.
- The URL helper should fail safely when the URL cannot be built.
- Tests were added to cover the new behavior.

## References

GEA-1809

## Checklist
- [x] Tests

### DIFF
--- a/agent_controller/agent_controller.c
+++ b/agent_controller/agent_controller.c
@@ -138,8 +138,8 @@ add_custom_header (gvm_http_headers_t *headers, const gchar *key,
  * The caller is responsible for freeing the allocated URL string with g_free().
  */
 static agent_controller_error_t
-agent_controller_set_url (agent_controller_connector_t conn, const gchar *path,
-                          gchar **url)
+agent_controller_build_url (agent_controller_connector_t conn,
+                            const gchar *path, gchar **url)
 {
   if (!conn)
     {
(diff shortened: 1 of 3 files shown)


agent_controller/agent_controller.c
rdttmr: This code block is technically a duplicate, `agent_controller_send_request_with_headers ()` already runs this check and produces this warning message.

No need to change anything, but as far as I am concerned this does not really do anything.
(`return NULL` also is the same as `return http_response` in this if-case)

agent_controller/agent_controller.c
```diff
@@ -88,57 +88,168 @@ init_custom_header (const gchar *apikey, gboolean content_type)
...
+ * @param[in] path The request path (e.g., "/api/v1/admin/agents").
+ * @param[out] url Pointer to a gchar* that will be allocated with the full URL.
+ *
+ * @return AGENT_CONTROLLER_OK on success, or an appropriate error code on
+ * failure (e.g., missing connection, missing URL components).
+ *
+ * The caller is responsible for freeing the allocated URL string with g_free().
+ */
+static agent_controller_error_t
+agent_controller_set_url (agent_controller_connector_t conn, const gchar *path,
```
rdttmr: I think the verb `set` does not do the function justice and could be a bit misleading :D

I personally would prefer something like
`agent_controller_build_url ()`
`agent_controller_make_url ()`
`agent_controller_construct_url ()`

however, leaving it as is also works for me.

robindittmar (dismissed):

robindittmar (approved):

PR #1078, opened 2026-06-05
Add: Add initial WAS wrapper client library.

a-h-abdelsalam:
## What
Add initial WAS wrapper client library.

## Why
Required for the integration of Web Application Scanner.

## References
GEA-1783

### DIFF
--- a/web_application_scanner/web_application_scanner.c
+++ b/web_application_scanner/web_application_scanner.c
@@ -76,7 +76,7 @@ add_credential_to_scan_json (gpointer credentials, gpointer cred_array)
 
   if (port)
     {
-      cJSON_AddNumberToObject (cred_obj, "port", atoi (port));
+      cJSON_AddNumberToObject (cred_obj, "port", strtol (port, NULL, 10));
     }
 
   cJSON *cred_type_obj = cJSON_CreateObject ();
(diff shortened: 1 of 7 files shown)


web_application_scanner/web_application_scanner.c
```diff
@@ -0,0 +1,241 @@
...
+ * @param urls           The URLs of the target.
+ * @param exclude_urls   The excluded URLs of the target.
+ *
+ * @return The newly allocated web_application_target_t.
+ */
+web_application_target_t *
+web_application_target_new (const gchar *scanid, const gchar *urls,
+                            const gchar *exclude_urls)
+{
+  web_application_target_t *new_target;
```
ozgen, line 207: `web_application_target_new()` currently allows creating a target without URLs. This can lead to a GLib critical warning or a crash when the builder is called with such a target.

I would prefer returning `NULL` if `urls` is `NULL`.

web_application_scanner/web_application_scanner.c
```diff
@@ -0,0 +1,241 @@
...
+  const gchar *type = scan_credential_get_type (cred);
+  const gchar *service = scan_credential_get_service (cred);
+  const gchar *port = scan_credential_get_port (cred);
+
+  cred_obj = cJSON_CreateObject ();
+  cJSON_AddStringToObject (cred_obj, "service", service ? service : "");
+
+  if (port)
+    {
+      cJSON_AddNumberToObject (cred_obj, "port", atoi (port));
```
ozgen: **Small concern:** maybe use `strtol()` instead of `atoi()`, because `atoi()` converts invalid ports to `0`.

ozgen (dismissed):

ozgen (approved):

PR #1087, opened 2026-07-06
Introducing timeout for "unix_vsendf_internal (...)".

jhelmold:
## What
Introduced a timeout for "unix_vsendf_internal (...)".

## Why
This fixes one scenario of the bug described in GEA-1642.

## References
GEA-1642

### DIFF
--- a/osp/osp.c
+++ b/osp/osp.c
@@ -147,9 +147,9 @@ osp_connection_new (const char *host, int port, const char *cacert,
       tv.tv_sec = 24;
       tv.tv_usec = 0;
       setsockopt(connection->socket, SOL_SOCKET, SO_RCVTIMEO, &tv,
-                 sizeof(struct timeval));
+                 sizeof (struct timeval));
       setsockopt(connection->socket, SOL_SOCKET, SO_SNDTIMEO, &tv,
-                 sizeof(struct timeval));
+                 sizeof (struct timeval));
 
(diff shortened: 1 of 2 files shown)


osp/osp.c
```diff
@@ -141,6 +141,16 @@ osp_connection_new (const char *host, int port, const char *cacert,
       memset (addr.sun_path, 0, sizeof (addr.sun_path));
       memcpy (addr.sun_path, host, strlen (host));
       len = strlen (addr.sun_path) + sizeof (addr.sun_family);
+
+      /* Set timeout */
+      struct timeval tv;
+      tv.tv_sec = 24;
```
timopollmeier, line 147: I don't think the timeout should be defined here.

Depending on how configurable we want to make this, it should be either:
* a `#define` constant
* a global variable
* a new parameter for `osp_connection_new`

util/serverutils.c
```diff
@@ -674,7 +676,7 @@ unix_vsendf_internal (int socket, const char *fmt, va_list ap, int quiet)
       count = write (socket, string, left);
       if (count < 0)
         {
-          if (errno == EINTR || errno == EAGAIN)
+          if ((errno == EINTR || errno == EAGAIN) && (timeout_count++ < 8))
```
timopollmeier, line 679: The retry limit should also not be hard-coded here with the same options as the timeout.

PR #1092, opened 2026-07-20
Change: Tweak and document logging

qha:
## What

Make syslog_facility and syslog_ident in logging configuration behave more intuitively.

Document logging configuration in readme.

Extend test should_load_log_configuration.

Extend the tests set in cmakelists.

Deduplicate and rearrange assert_thats in test should_convert_facility_int_from_string.

## Why

To pass syslog_facility according to configuration for all log domains, even those not explicitly named in configuration. (At present the compiled in default of "local0" supersedes the configuration for log domain "*".)

To pass ident to openlog for all log domains, even those not explicitly named in configuration.

To set expectations and lower the surprise factor.

To test logging configuration more thoroughly.

To make `make tests` compile/link all tests that `make test` expects to find.

To make the test should_convert_facility_int_from_string marginally easier to read.

## Checklist

- [x] Tests

## Notes for reviewers

I'd be happy to get feedback, in particular on the correctness and readability of the new readme section and if the changes of behaviour in "Change: Remove syslog_facility fallback in load_log_configuration" and "Change: Make syslog_ident default to log_domain" are big enough to warrant the change prefix.

I'd also like to note that C programming is not my forte so please review carefully.

I've made an effort to ensure that everything i documented is tested. Please let me know if you see something i missed.

### DIFF
--- a/base/logging_tests.c
+++ b/base/logging_tests.c
@@ -134,14 +134,20 @@ Ensure (logging, should_load_log_configuration)
   /* Create a temporary configuration file */
   FILE *file = fopen (config_file, "w");
   assert_that (file, is_not_null);
-  fprintf (file, "[*]\n"
+  fprintf (file, "[bar]\n"
+                 "file=syslog\n"
+                 "syslog_facility=kern\n"
+                 "syslog_ident=test_ident\n"
+                 "\n"
+                 "[foo]\n"
+                 "level=debug\n"
+                 "\n"
+                 "[*]\n"
                  "prepend=%%t %%s %%p - \n"
                  "separator=:\n"
                  "prepend_time_format=%%Y-%%m-%%d %%H:%%M:%%S\n"
                  "file=-\n"
-                 "level=debug\n"
-                 "syslog_facility=local0\n"
-                 "syslog_ident=test_ident\n");
+                 "level=info\n");
   fclose (file);
 
   /* Load the configuration */
@@ -151,6 +157,8 @@ Ensure (logging, should_load_log_configuration)
   /* Verify the configuration */
   gvm_logging_domain_t *log_domain_entry =
     (gvm_logging_domain_t *) log_config_list->data;
+  assert_that (gvm_logging_domain_get_log_domain (log_domain_entry),
+               is_equal_to_string("*"));
   assert_that (gvm_logging_domain_get_prepend_string (log_domain_entry),
                is_equal_to_string ("%t %s %p - "));
   assert_that (gvm_logging_domain_get_prepend_separator (log_domain_entry),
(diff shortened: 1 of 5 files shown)


base/logging_tests.c
```diff
@@ -159,13 +167,58 @@ Ensure (logging, should_load_log_configuration)
...
+  assert_that (gvm_logging_domain_get_syslog_facility (log_domain_entry),
+               is_equal_to_string ("kern"));
   assert_that (gvm_logging_domain_get_syslog_ident (log_domain_entry),
                is_equal_to_string ("test_ident"));
 
+  log_config_list = g_slist_next (log_config_list);
+  assert_that (log_config_list, is_null);
+
   /* Clean up */
   free_log_configuration (log_config_list);
```
ozgen, line 221: I think we should keep the original head of `log_config_list` here. Since we iterate through the list until it becomes `NULL`, `free_log_configuration(log_config_list)` below doesn't actually free the loaded configuration.

Maybe use a separate iterator, e.g. `log_config_entry`, while keeping `log_config_list` for cleanup.

qha: Good catch, thanks! I will get back with a fix for this.

qha: Pushed another commit. Let me know if you'd like me to squash it into "Test: Extend test should_load_log_configuration".

ozgen (dismissed):

ozgen:
@qha, could you please fix the formatting issues by running:

`clang-format -i -style=file {agent_controller,base,boreas,gmp,http,http_scanner,openvasd,osp,util,security_intelligence,web_application_scanner}/*.{c,h}`

qha:
Certainly, please see the commit i just pushed.

ozgen (dismissed):

ozgen:
> Certainly, please see the commit i just pushed.

The PR still has some formatting issues, @qha. Could you please run the `clang-format` command I shared earlier?

qha:
Your exact command made no changes. I expanded the list of directories and got some changes to the container_image_scanner, cyberark and ldap directories. I added a commit to cover them to in the command run by ci, i don't usually work with Github workflows so please take some extra care in reviewing that one!

qha:
(I've tried now with both:
```console
% clang-format --version
Debian clang-format version 19.1.7 (3+b1)
```
and:
```console
% clang-format --version
Ubuntu clang-format version 21.1.8 (6ubuntu1)
```
so i really hope the directory list turns out to have been the problem.)

ozgen (approved):

ozgen:
> (I've tried now with both:
> 
> ```
> % clang-format --version
> Debian clang-format version 19.1.7 (3+b1)
> ```
> 
> and:
> 
> ```
> % clang-format --version
> Ubuntu clang-format version 21.1.8 (6ubuntu1)
> ```
> 
> so i really hope the directory list turns out to have been the problem.)

I checked out your branch and fixed the remaining formatting issues. Thanks for your contributions!

qha:
My pleasure and thanks for your help!

PR #1118, opened 2026-08-31
Fix gvm_get_outgoing_iface NULL deref and fd left open

timopollmeier:
## What
In the `gvm_get_outgoing_iface` function, `target_add` is now checked for NULL before it is dereferenced.
The socket file descriptor opened in the function is also closed.

## References
GEA-1985

### DIFF
--- a/base/networking.c
+++ b/base/networking.c
@@ -1300,11 +1300,11 @@ gvm_get_outgoing_iface (struct sockaddr_storage *target_addr)
   struct sockaddr_storage out_iface_addr;
   char *out_iface_str;
 
-  out_iface_str = NULL;
-  family = target_addr->ss_family;
-
   if (!target_addr)
     return NULL;
+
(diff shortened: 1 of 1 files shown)


a-h-abdelsalam (approved):

base/networking.c
```diff
@@ -1300,21 +1300,25 @@ gvm_get_outgoing_iface (struct sockaddr_storage *target_addr)
...
   // get a connected udp socket
   sockfd = get_connected_udp_sock (target_addr);
   if (sockfd < 0)
     return NULL;
   // get socked address which is the addr of the interface we want to get
   out_iface_addr.ss_family = family;
   if (get_sock_addr (sockfd, &out_iface_addr) < 0)
-    return NULL;
+    {
+      close (sockfd);
```
mattmundell, line 1317: `get_sock_addr` is already closing on error.

PR #1120, opened 2026-09-03
add: support plain support bundle downloads

ozgen:
## What

* Add support for downloading decrypted support bundles from the Agent Controller.
* Add support for the optional `days` parameter.
* Add unit tests for the new plain support bundle download function.

## Why

* Allow gvmd to retrieve support bundles without encryption.
* Support the new Agent Controller `/support-bundle/plain` endpoint.

## References

GEA-2040

## Checklist

- [x] Tests

### DIFF
--- a/agent_controller/agent_controller.c
+++ b/agent_controller/agent_controller.c
@@ -2175,3 +2175,103 @@ agent_controller_download_support_bundle (agent_controller_connector_t conn,
 
   return bundle;
 }
+
+/**
+ * @brief Downloads a decrypted support bundle for a specific agent.
+ *
+ * @param[in] conn Active connector to the Agent Controller.
+ * @param[in] agent_id ID of the agent.
+ * @param[in] days Number of days of logs to include. A value of 0 uses the
+ *                 Agent Controller's configured default.
+ *
+ * @return Newly allocated support bundle on success, NULL on failure.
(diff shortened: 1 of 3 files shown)


agent_controller/agent_controller.c
```diff
@@ -2175,3 +2175,103 @@ agent_controller_download_support_bundle (agent_controller_connector_t conn,
...
+ * @param[in] conn Active connector to the Agent Controller.
+ * @param[in] agent_id ID of the agent.
+ * @param[in] days Number of days of logs to include. A value of 0 uses the
+ *                 Agent Controller's configured default.
+ *
+ * @return Newly allocated support bundle on success, NULL on failure.
+ *         Free with agent_controller_support_bundle_free().
+ */
+agent_controller_support_bundle_t
+agent_controller_download_support_bundle_plain (
```
a-h-abdelsalam, line 2225: There seems to be some code duplication between this and the function above. May be you could refactor the shared logic into a private helper function.

ozgen: fixed with: https://github.com/greenbone/gvm-libs/pull/1120/commits/e0277304b45289578bcce2b0ba395fcb73c42bf3

a-h-abdelsalam (approved):

PR #1121, opened 2026-09-07
Add `gvm_is_valid_xml` function

timopollmeier:
## What
A new function to check if a string is valid XML is added to `util/xmlutils`.

## Why
This is intended for validating XML file uploads in GSA

## References
GEA-2036

## Checklist
- [x] Tests

### DIFF
--- a/util/xmlutils.c
+++ b/util/xmlutils.c
@@ -3058,3 +3058,40 @@ xml_file_iterator_next (xml_file_iterator_t iterator, gchar **error)
 
   return NULL;
 }
+
+/**
+ * @brief Check if a string is valid XML.
+ *
+ * @param[in] str  The string to check.
+ */
+int
+gvm_is_valid_xml (const char *str, gchar **error_message)
+{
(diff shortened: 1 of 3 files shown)


util/xmlutils.c
```diff
@@ -3058,3 +3058,40 @@ xml_file_iterator_next (xml_file_iterator_t iterator, gchar **error)
...
+
+  memset (&sax_handler, 0, sizeof (xmlSAXHandler));
+  sax_handler.initialized = XML_SAX2_MAGIC;
+  if (error_message)
+    *error_message = NULL;
+
+  xmlParserCtxt *ctx =
+    xmlCreatePushParserCtxt (&sax_handler, NULL, NULL, 0, NULL);
+
+  ret = xmlParseChunk (ctx, str, strlen (str), 1);
```
ozgen, line 3092: Should we handle str == NULL here for `strlen (str)`?

ozgen (approved):

PR #1123, opened 2026-09-15
Add: Add a new ldap library

a-h-abdelsalam:
## What
Added a new ldap library that acts as a high-level wrapper around OpenLDAP and supports simple bind and search operations.

## Why
This will be used in integrating with LDAP directories.

## References
GEA-2056

## Checklist

- [x] I have used the following LLMs/AI tools in this pull request: GPT-5.3-Codex/GPT-5.5


ldap/gvmldap.h
```diff
@@ -0,0 +1,99 @@
...
+                       gvm_ldap_search_callback_t, gpointer);
+
+void
+gvm_ldap_close (LDAP *);
+
+gchar *
+gvm_ldap_entry_get_dn (LDAP *, LDAPMessage *);
+
+gchar *
+gvm_ldap_entry_get_string (LDAP *, LDAPMessage *, const gchar *);
```
ozgen: Do we intentionally want to expose `LDAP` and `LDAPMessage` in the public API? If this library is meant to abstract `OpenLDAP`, it might be better to hide the `OpenLDAP` types. Otherwise consumers still depend directly on the `OpenLDAP` API.

a-h-abdelsalam: Thanks! I updated the PR.

ozgen (approved):

PR #1125, opened 2026-09-21
Doc: Adjustments

qha:
## What

Adjust docs.

## Why

To keep the docs up to date with changes in this project and in the environment.

## Checklist


INSTALL.md
```diff
@@ -17,17 +17,17 @@ General build environment:
 
 Specific development libraries:
 
-* libcjson >= 1.7.14 (util)
-* libcurl >= 7.83.0 (openvasd)
-* libgcrypt
+* libcjson >= 1.7.14 (agent_controller, container_image_scanner, cyberark, http_scanner, openvasd, security_intelligence, util, web_application_scanner)
+* libcurl >= 7.83.0 (agent_controller, container_image_scanner, cyberark, http, http_scanner, openvasd, security_intelligence)
+* libgcrypt (util)
```
ozgen, line 22: Could you add ldab library in here:
`libldap >= 2.4 (ldab)`

INSTALL.md
```diff
@@ -69,17 +69,17 @@ Certain features of gvm-libs are optional and require the following:
 
 Prerequisites for LDAP authentication:
 
-* libldap2 library >= 2.4.44 (util) (Debian package: libldap2-dev)
+* libldap library >= 2.4.44 (util, ldap) (Debian package: libldap-dev)
```
ozgen: should LDAP still be listed under optional prerequisites? The CMake check currently requires ldap >= 2.4, so configuration fails if the package is missing. If LDAP is now required for the default build, I think we should move it to the regular prerequisites as well.

INSTALL.md
qha: Certainly! That was actually what i wanted to do when i began this pull request but then i found it under optional and thought perhaps that's where it should be as the dependency can be turned off. Please see the new commit i pushed.

INSTALL.md
qha: Agreed, please see my comment in your other conversation and the new commit i pushed.

qha:
I'm logging out for today now so if you have other comments i'll look at them tomorrow.

ozgen (approved):

PR #1126, opened 2026-09-23
Add: Add a new AD connector.

a-h-abdelsalam:
## What
Added a new AD connector using `gvmldap`.

## Why
Supports Active Directory connection and searches.

## References
GEA-2079

## Checklist

- [X] I have used the following LLMs/AI tools in this pull request:


ad_connector/ad_connector.c
```diff
@@ -0,0 +1,1059 @@
...
+    case AD_CONNECTOR_SEARCH_SCOPE_BASE:
+      return GVM_LDAP_SCOPE_BASE;
+    case AD_CONNECTOR_SEARCH_SCOPE_ONELEVEL:
+      return GVM_LDAP_SCOPE_ONELEVEL;
+    case AD_CONNECTOR_SEARCH_SCOPE_SUBTREE:
+      return GVM_LDAP_SCOPE_SUBTREE;
+    default:
+      g_warning ("%s: Unknown filter scope %d. Defaulting to onelevel scope.",
+                 __func__, scope);
+      return GVM_LDAP_SCOPE_ONELEVEL;
```
ozgen: I don't think silently changing an invalid enum into a valid search scope is ideal. It can hide programming errors.

ad_connector/ad_connector.h
```diff
@@ -0,0 +1,152 @@
...
+
+ad_connector_return_t
+ad_connector_connect (ad_connector_t, const gchar *);
+
+ad_connector_return_t
+ad_connector_search_objects (ad_connector_t, const ad_search_config_t,
+                             ad_object_callback_t, gpointer,
+                             ad_object_search_result_t *);
+
+#endif /* not _GVM_AD_CONNECTOR_AD_CONNECTOR_H */
```
ozgen: add 1 extra line in the end

ad_connector/ad_connector.c
```diff
@@ -0,0 +1,1059 @@
...
+  if (!attributes)
+    {
+      g_warning ("%s: Failed to build LDAP attributes list.", __func__);
+      g_free (filter);
+      return AD_CONNECTOR_INVALID_VALUE;
+    }
+
+  gvm_ldap_scope_t scope = get_ldap_search_scope (search_config->scope);
+
+  search_params = gvm_ldap_search_params_new (
```
ozgen: `ad_search_config_new()` leaves `base_dn` unset. When search eventually happens, this gets passed it directly.  Probably the lower-level LDAP implementation rejects a NULL base DN.
Not a blocker, I just confused unclear ownership of validation.

ad_connector/ad_connector.c
```diff
@@ -0,0 +1,1059 @@
...
+      ad_object_free (object);
+      return GVM_LDAP_SEARCH_STOP;
+    }
+
+  result = ctx->callback (object, ctx->user_data);
+  ad_object_free (object);
+
+  switch (result)
+    {
+    case AD_OBJECT_CALLBACK_CONTINUE:
```
ozgen: The callback ownership design looks good. Would it make sense to also support a successful `STOP` result?

For example:

```c
if (found_what_we_need)
  return AD_OBJECT_CALLBACK_STOP;
```

ad_connector/ad_connector_tests.c
```diff
@@ -0,0 +1,1566 @@
...
+
+  if (argc > 1)
+    ret = run_single_test (suite, argv[1], create_text_reporter ());
+  else
+    ret = run_test_suite (suite, create_text_reporter ());
+
+  destroy_test_suite (suite);
+
+  return ret;
+}
```
ozgen: Could you add one extra blank line at the end?

ad_connector/ad_connector.c
```diff
@@ -0,0 +1,1059 @@
...
+    {
+      gvm_ldap_close (connector->ldap_connection);
+    }
+
+  g_free (connector->ldap_host);
+  g_free (connector->cacert_file);
+  g_free (connector->bind_dn);
+
+  g_free (connector);
+}
```
ozgen: Could you add one extra blank line at the end?

ad_connector/ad_connector.c
a-h-abdelsalam: `ad_search_config_builder` already rejects invalid `base_dn` but I also added a validation in `ad_connector_search_objects`.

ad_connector/ad_connector.c
a-h-abdelsalam: yes makes sense to support a normal successful stop.

ad_connector/ad_connector.c
a-h-abdelsalam: I updated it to fail for invalid values

ozgen (approved):

PR #1127, opened 2026-09-23
Change: Validate attribute names when building new search parameters.

a-h-abdelsalam:
## What
Validate attribute names when building new search parameters.

## Why
Was missing.

## References
GEA-2079

## Checklist

- [X] I have used the following LLMs/AI tools in this pull request:


ldap/gvmldap.c
```diff
@@ -389,13 +389,51 @@ gvm_ldap_scope_to_openldap (gvm_ldap_scope_t scope, int *ldap_scope)
...
+ * @details The attribute name must start with an ASCII letter and
+ * can be followed by zero or more ASCII letters, digits,
+ * or hyphens.
+ *
+ * @param attribute LDAP attribute name to validate.
+ *
+ * @return TRUE if the attribute name is valid, FALSE otherwise.
+ */
+static gboolean
+gvm_ldap_attribute_name_is_valid (const gchar *attribute)
```
ozgen: I think this validation looks too restrictive.

LDAP attribute descriptions can use numeric OIDs and may include options. For example:

```text
2.5.4.0
cn;lang-de
```

RFC 4512 defines this `AttributeDescription` format:
https://www.rfc-editor.org/rfc/rfc4512.html

Microsoft  AD documentation also references numeric OIDs and `AttributeDescription` options:
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/7cda533e-d7a4-4aec-a517-91d02ff4a1aa

Should this validation allow these forms as well?

a-h-abdelsalam: yes I agree and I had this in mind and had already implemented validation for `OIDs` and options but then decided to take them out. 
The reason is that the required attributes are hard-coded for now in `ad_connector` and any extra attributes (if needed) will probably be hard-coded in `gvmd` for a specific detailed view. So we can decide what format we want to use.

ozgen (approved):

